PR #29, opened 2018-11-26
Installs git precommit hook for testing before committing

jasonharrison:
Installs git precommit hook using https://www.npmjs.com/package/pre-commit

Tests before committing (use --no-verify to override).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,9 @@
     "dev": "tsc-watch --onSuccess \"npm run webpack-run\"",
     "precommit": "npm run build && npm run test && npm run lint"
   },
+  "pre-commit": [
+    "precommit"
+  ],
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,6 +15,9 @@
     "dev": "tsc-watch --onSuccess \"npm run webpack-run\"",
     "precommit": "npm run build && npm run test && npm run lint"
   },
+  "pre-commit": [
+    "precommit"
```
flamecoals: you probably want to replace this with:
"test"
"lint"

jasonharrison: It runs ```npm run precommit``` 

Which is equivalent to ``` npm run build && npm run test && npm run lint```

https://github.com/flamecoals/turnato/blob/5fba8c5824e437c349080a6f3d69d77516645001/package.json#L16

Spongebob Meme: MAGIC!

PR #45, opened 2019-01-16
Blink <SvgExplosion> or <SvgMiss> after firing a salvo

jasonharrison:
## Known issues:

* On page refresh all salvos will blink
* Add tests?


flamecoals:
Thanks for this PR! I will have a look into it tomorrow.

PR #68, opened 2019-01-27
[WIP] Notifications to alert users when it is their turn

### LINKED ISSUE #61: [Infrastructure] Automatic notifications for the user when it is their turn so they can play.
### Requisites

- If the user is already seing the screen, no need to beep/notify them.

jasonharrison:
This PR will close #61


src/games/seabattle/Battle.tsx
```diff
@@ -89,6 +95,18 @@ export class Battle extends React.Component<IBattleProps, IBattleState> {
...
   }
 
+  _notifyPlayer() {
+    if (this.state.playerID === this.state.currentPlayer) {
+      if (!this.state.playerNotified) {
+        this.setState({
+          ...this.state,
+          playerNotified: true,
+        });
+        return (<Notification title={'Your Turn!'} />);
```
flamecoals: this will only work if the user has the tab open, which will not be the case for mobile users. We need to do server-side push notifications for mobile, but this can be a temporary solution

flamecoals:
Can you add a test case (as documentation) for when you expect the notification to be fired? Besides that, LGTM.

src/games/seabattle/Battle.tsx
jasonharrison: That's what I thought.  I know you said this is low-priority, so we can revisit this later.

src/games/seabattle/Battle.test.tsx
```diff
@@ -75,3 +75,39 @@ test('one phase - miss', () => {
...
+  expect(grid.find('Notification').length).equals(0);
+  client.updatePlayerID('0');
+  const grid2 = Enzyme.mount(
+    <Battle
+      G={store.getState().G}
+      moves={client.moves}
+      playerID={'0'}
+      currentPlayer={store.getState().ctx.currentPlayer}
+    />);
+  expect(grid.find('Notification').length).equals(0);
```
flamecoals: can you test when a notification happens, not only when it doesnt happen?

jasonharrison:
Todo: 
* Notifications still occur even on invite screen
* Notifying users while on the page?

image

flamecoals:
you can do this:
https://coderwall.com/p/cwpdaw/how-to-tell-if-app-page-has-focus

add these event listener when the component is constructed, and remove them when the component is destroyed.

on the callback, set the state of the component to either focused or not.

if the component is NOT focused, render the Notification component.

flamecoals:
also check that the invite screen is not open :)

jasonharrison:
I am stuck with this because GameBoardWrapper only seems to render once.  

Maybe we can do this through an enhancer?

jasonharrison:
Putting this on pause until the lobby is finished because this PR will require the use of a server component to be useful.

flamecoals:
This is old now, and I dont think this is a high priority right now, until we have a good database after lobby we can revisit server-side notifications.

PR #76, opened 2019-01-30
Replace sample text in header

jasonharrison:
image

@flamecoals you said you were thinking of updating this text?

### DIFF
--- a/src/Home/Header.tsx
+++ b/src/Home/Header.tsx
@@ -8,14 +8,8 @@ class Header extends React.Component<{}, {}> {
         <CardTitle title="Free as in freedom" />
         <CardText>
           Play when you can, wherever you are.
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit,
-          sed do eiusmod tempor incididunt ut labore et dolore magna
-          aliqua. Ut enim ad minim veniam, quis nostrud exercitation
-          ullamco laboris nisi ut aliquip ex ea commodo consequat.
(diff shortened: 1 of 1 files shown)


jasonharrison:
@flamecoals Here are two designs:

**With card:**

image

---

**If we get rid of the card:**

image

---

I think I like the latter more.  Thoughts?

flamecoals:
I like the latter more too

src/Home/Header.tsx
```diff
@@ -8,14 +8,8 @@ class Header extends React.Component<{}, {}> {
...
-          Lorem ipsum dolor sit amet, consectetur adipiscing elit,
-          sed do eiusmod tempor incididunt ut labore et dolore magna
-          aliqua. Ut enim ad minim veniam, quis nostrud exercitation
-          ullamco laboris nisi ut aliquip ex ea commodo consequat.
-          Duis aute irure dolor in reprehenderit in voluptate velit
-          esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
-          occaecat cupidatat non proident, sunt in culpa qui officia
-          deserunt mollit anim id est laborum.
+          Turnato is a free (as in freedom), mobile-first, board game platform.
+          Its goal is to popularize board games and to make them easy to play with friends, even from afar.
```
flamecoals: The text you are showing on the screenshot is not here. Can you update?

Also, I think we dont need the same card on the "about" page. Maybe just links to github and etc.

PR #78, opened 2019-02-03
Fix #77 - Add Windows support to "yarn run dev"

### LINKED ISSUE #77: Error: ENOENT: no such file or directory, open './dist/webpack/template.html'
I tried to clone this today and get it running but was unable to run `yarn run dev`. I was able to get it to run using `yarn run prod`, though. For context, I'm on Windows. Did I skip a step?

Repro steps:

* Clone repo
* CD into repo
* `yarn install`
* `yarn run dev`

**Output:**
```
$ webpack --progress --config dev.webpack.config.js; tsc; webpack --watch --progress --config dev.webpack.config.js | node dist/server.js
fs.js:119
    throw err;
    ^

Error: ENOENT: no such file or directory, open './dist/webpack/template.html'
```

hoovercj:
Fixes #77 

"yarn run dev" does not work on Windows as the command separators are not properly recognized. Changing them all to "&&" as they are in the "prod" script allows this command to also be run on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "test:coveralls": "cat coverage/lcov.info | node node_modules/.bin/coveralls",
     "build": "tsc",
     "prod": "yarn run build && webpack --config prod.webpack.config.js && pm2 start pm2-prod.config.js",
-    "dev": "webpack --progress --config dev.webpack.config.js; tsc; webpack --watch --progress --config dev.webpack.config.js | node dist/server.js",
+    "dev": "webpack --progress --config dev.webpack.config.js && tsc && webpack --watch --progress --config dev.webpack.config.js && node dist/server.js",
     "precommit": "yarn run verify-lint && yarn run verify-tsfmt && yarn run build && yarn run test",
     "pre": "yarn run precommit",
     "fix": "yarn run lint && yarn run tsfmt",


package.json
```diff
@@ -10,7 +10,7 @@
     "test:coveralls": "cat coverage/lcov.info | node node_modules/.bin/coveralls",
     "build": "tsc",
     "prod": "yarn run build && webpack --config prod.webpack.config.js && pm2 start pm2-prod.config.js",
-    "dev": "webpack --progress --config dev.webpack.config.js; tsc; webpack --watch --progress --config dev.webpack.config.js | node dist/server.js",
+    "dev": "webpack --progress --config dev.webpack.config.js && tsc && webpack --watch --progress --config dev.webpack.config.js && node dist/server.js",
```
flamecoals: this does not work in ubuntu for me :(. If we change the pipe to &&, the server would only run after webpack is complete, and because it is in watch mode it is never complete.

Does it break windows if we keep the pipe? I will boot up my windows machine and test.

flamecoals: btw, I think there _must_ be a better way to do this, it seems very hacky right now. I will have a look on how create-react-app makes it work.

jasonharrison: I was working on a better solution that uses npm-run-all instead of pipes and &&, but I didn't get the chance to test.

flamecoals: I tested in my windows machine "webpack --progress --config dev.webpack.config.js && tsc && webpack --watch --progress --config dev.webpack.config.js | node dist/server.js" (with the pipe) and it worked ! I will commit it as a stop-gap solution. Thanks @hoovercj for helping :)

PR #93, opened 2019-02-17
Enabling first code splitting

flamecoals:


server-build/37a678d8328ecfa585ff.js
```diff
@@ -0,0 +1,76 @@
+exports.ids = [0];
```
jasonharrison: This probably shouldn't be here

flamecoals: Indeed, thanks for catching. Done

PR #94, opened 2019-02-18
Code splitting for each game

flamecoals:


flamecoals:
still WIP, need to fix styling after upgrading material-ui and get 100% test coverage.

PR #114, opened 2019-02-20
Remove robots.txt while building Docker image

jasonharrison:
Remove robots.txt while building our Docker image.  This way, yarn run dev will have a /robots.txt that disallows crawling, but our production docker images will not.


jasonharrison:
Actually I think i'm going to deal with this another way.

PR #120, opened 2019-02-21
Homepage links

jasonharrison:
* Add link to Discord
* Move many links up on the homepage as material-ui logos

### DIFF
--- a/src/Home/Header.tsx
+++ b/src/Home/Header.tsx
@@ -20,7 +20,7 @@ class Header extends React.Component<{}, {}> {
           Not only do you get to enjoy free high quality games everywhere,
           but you also can study how they are made, change them, and contribute back to the community!
         </Typography>
-        _links();
+        {this._links()}
       </div>
     );
   }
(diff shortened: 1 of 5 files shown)


src/Home/Header.tsx
```diff
@@ -14,6 +20,32 @@ class Header extends React.Component<{}, {}> {
           Not only do you get to enjoy free high quality games everywhere,
           but you also can study how they are made, change them, and contribute back to the community!
         </Typography>
+        _links();
```
flamecoals: Does this work  ? I thought you would need to do this._links();

flamecoals: _links() is NOT being interpreted as code. This is being interpreted as text. Thats why coverage decreased. You need to use {this._links()} instead.

jasonharrison: Yep -- I already fixed this in my local copy. I didn't push yet because I got sidetracked by our server problem. I'll work on it tomorrow (well.. it's 3:15 am, so I guess today).

PR #125, opened 2019-02-22
MessagePage: add user-friendly link to go home in the case of an error or a loading timeout (loading for > 5 seconds)

jasonharrison:
Working on MessagePage to have a more obvious link to go home in case of an error or a loading timeout (loading for > 5 seconds).

### DIFF
--- a/src/App/MessagePageClass.test.tsx
+++ b/src/App/MessagePageClass.test.tsx
@@ -12,14 +12,14 @@ test('MessagePage starts with link hidden', () => {
   const msg = Enzyme.mount(
     <MemoryRouter><MessagePage type={'loading'} message={'loading'} /></MemoryRouter>,
   ).find(MessagePage);
-  expect(msg.html()).not.to.contain('Go home');
+  expect(msg.html()).not.to.contain('Go Home');
 });
 
 test('MessagePage starts with link visible for errors', () => {
   const msg = Enzyme.mount(
     <MemoryRouter><MessagePage type={'error'} message={'Not Found'} /></MemoryRouter>,
   ).find(MessagePage);
-  expect(msg.html()).to.contain('Go home');
+  expect(msg.html()).to.contain('Go Home');
 });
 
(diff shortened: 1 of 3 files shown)


jasonharrison:
# Error:
image

# Loading (before 5 seconds):
image


# Loading (after 5 seconds):
image

flamecoals:
Great job!

src/App/MessagePageClass.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+test('MessagePage starts with link visible for errors', () => {
+  const msg = Enzyme.mount(
+    <MemoryRouter><MessagePage type={'error'} message={'Not Found'} /></MemoryRouter>,
+  ).find(MessagePage);
+  expect(msg.html()).to.contain('Go Home');
+});
+
+test('MessagePage is hidden', () => {
```
flamecoals: The link is hidden, not the page

src/App/MessagePageClass.tsx
```diff
@@ -0,0 +1,76 @@
...
+    };
+    if (props.type !== 'error') {
+      requestAnimationFrame(this._animate(Date.now()));
+    }
+  }
+
+  _animate = (now: number) => () => {
+    const elapsed = now - this.state.startTime;
+    const timedout = elapsed > 5000;
+    const linkHidden = timedout ? false : this.props.type !== 'error';
```
flamecoals: I don't think you need to check the type here given that _animate won't be called in case of errors. So in effect link hidden === timed out, you can even remove the link hidden variable.

src/App/MessagePageClass.test.tsx
jasonharrison: Oops, thanks.

src/App/MessagePageClass.tsx
jasonharrison: You're right, thank you!

PR #149, opened 2019-03-06
 Serve different robots.txt depending on the request host

jasonharrison:
Serve different robots.txt depending on the request host


src/server.tsx
```diff
@@ -73,11 +81,19 @@ const startServer = async () => {
...
     await next();
-    const { render, status } = await renderSite(ctx.request.url);
-    if (status) {
-      ctx.response.status = Number(status);
+    if (ctx.request.path === '/robots.txt') {
+      if (ctx.request.hostname.toLowerCase() === 'freeboardgame.org') {
+        ctx.response.body = OPEN_ROBOTS_TXT;
+      } else {
+        ctx.response.body = RESTRICTIVE_ROBOTS_TXT;
+      }
```
flamecoals: can you move this logic of rendering the robot txt to its own method? I think it is a little bit too much here in startServer

src/server.tsx
```diff
@@ -23,6 +23,14 @@ const NODE_ENV = process.env.NODE_ENV;
 const PROD = NODE_ENV === 'production';
 const DEV = !PROD;
 
+const RESTRICTIVE_ROBOTS_TXT = `User-agent: *
+Disallow: /`;
+
+const OPEN_ROBOTS_TXT = `User-agent: *
+Disallow: /g/chess/online/*
+Disallow: /g/seabattle/online/*
```
flamecoals: We have access to the GAMES_LIST constant. Can we generate this based on that constant? So we dont need to hard code values

src/server.tsx
jasonharrison: I agree, changed

PR #154, opened 2019-03-07
Fixing castling bug for chess

flamecoals:


src/games/chess/board.test.tsx
```diff
@@ -234,3 +234,39 @@ test('little AI game', () => {
   expect(moveMock.mock.calls[0]).to.deep.equal(['f4']);
   expect(stepMock.mock.calls.length).to.equal(1);
 });
+
+test('castling fix', () => {
+  const board = Enzyme.mount((
```
jasonharrison: I like that you added a test for this.

PR #163, opened 2019-03-10
Seabattle AI

### LINKED ISSUE #53: Seabattle: Add single-player mode against bot

jasonharrison:
Seabattle AI closes #53


src/games/seabattle/Battle.tsx
```diff
@@ -40,6 +42,9 @@ export class Battle extends React.Component<IBattleProps, IBattleState> {
     ).length === 0;
     if (uniqueMove) {
       this.props.moves.salvo(cell.x, cell.y);
+      if (this.props.isAIGame) {
+        setTimeout(this.props.step, 3000);
```
flamecoals: why 3s ? isnt it too much? cant it be 1s ?

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+    return { action: { type: 'MAKE_MOVE', payload: { type: 'setShips', args: [ships], playerID: '1' } } };
+  }
+
+  makeSalvoMove(cell: ICell) {
+    return { action: { type: 'MAKE_MOVE', payload: { type: 'salvo', args: [cell.x, cell.y], playerID: '1' } } };
+  }
+
+  generateMove(state: IPlayState) {
+    const salvos = state.G.salvos.filter((salvo: ISalvo) => (
+      salvo.player === 1 &&
```
flamecoals: this salvo.player === 1 should probably not be hard coded... you have the playerID coming from play() and you could pass into the generateMove

flamecoals: otherwise the Simulate wont work

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+        // no other salvos for the same ship were hit, so hit a random neighbor
+        return this.getRandomNeighbor(state, salvo);
+      }
+    }
+    // generate a random move
+    return this.generateRandomMove(state);
+  }
+
+  getMoveFromOtherSalvos(state: IPlayState, hitSalvos: ISalvo[]): ICell {
```
flamecoals: can we rename this to be "getNextShipFoundMove"

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+        if (move.y === 0) {
+          move.y++;
+        } else if (move.y === 9) {
+          move.y--;
+        } else {
+          move.y += randomChange;
+        }
+      }
+      validMove = this.isValidMove(state, move);
+      i++;
```
flamecoals: this means that you can return an undefined move, if i gets to 10. How are you treating this case?

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+        } else if (move.x === 9) {
+          move.x--;
+        } else {
+          move.x += randomChange;
+        }
+      } else {
+        if (move.y === 0) {
+          move.y++;
+        } else if (move.y === 9) {
+          move.y--;
```
flamecoals: you dont need to check the bounds here, given that if it is out of bounds the isValidMove will return false and it will try again.

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+    let i = 0;
+    let vertical;
+    let validMove;
+    let randomChange;
+    let move: ICell;
+    while (!validMove && i < 10) {
+      vertical = this.genRandomNumber(0, 1) === 0;
+      move = { x: salvo.cell.x, y: salvo.cell.y };
+      randomChange = (this.genRandomNumber(0, 1) === 0) ? -1 : 1;
+      if (vertical) {
```
flamecoals: this would look simpler:
moveDelta = { x: (vertical) ? randomChange : 0, y: (vertical) ? 0 : randomChange };
move = { x: salvo.cell.x + moveDelta, y: salvo.cell.y + moveDelta }
Given that you dont need to check for the bounds (it is done in isValidMove), this should be enough

src/games/seabattle/ai.ts
flamecoals: Also, I think the i should be much larger, maybe 100, given that there are some randomness here. 
Given that we know that the ship is not sunk, for sure there is a neighbor, so we dont need the i check.

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+    // generate a random move
+    return this.generateRandomMove(state);
+  }
+
+  getMoveFromOtherSalvos(state: IPlayState, hitSalvos: ISalvo[]): ICell {
+    const len = hitSalvos.length;
+    const lastHitSalvo = hitSalvos[len - 1];
+    const lastHitCell = lastHitSalvo.cell;
+    const secondLastHitCell = hitSalvos[len - 2].cell;  // second to last hit
```
flamecoals: I dont like that you are relying on the order of the salvos, doesnt make sense to me.
Why would you instead do this:
```
minSalvoPos = { x: (min of all x in hitSalvos), y: (min of all y in hitSalvos) }
maxSalvoPos = { x: (max of all x in hitSalvos), y: (max of all y in hitSalvos) }
diffMinMaxSalvoPos = { x: minSalvoPos.x - maxSalvoPos.x, y: minSalvoPos.y - maxSalvoPos.y }  // either x or y must be 0, given that the ship must be horizontal or vertical
direction = diffMinMaxSalvoPos.x === 0 ? { x: 0, y: 1 } :  { x: 1, y: 0 };
diffLength = diffMinMaxSalvoPos.x + diffMinMaxSalvoPos.y
if (diffLength == salvos.length) { // This means that there is no "hole" in the salvos, therefore we must try the edges
  return anyValidMove([{ x: minSalvoPos.x - direction.x, y: minSalvoPos.y - direction.y }, { x: maxSalvoPos.x + direction.x, y: maxSalvoPos.y + direction.y }]); 
} else {
  return anyValidMove([ all cells between minSalvoPos and maxSalvoPos ]);
}
```

src/games/seabattle/ai.ts
jasonharrison: I actually noticed this the other day, fixed it in TicTacToe's AI, and intended to fix in Seabattle but forgot.  https://github.com/freeboardgame/FreeBoardGame.org/pull/175 :-p

src/games/seabattle/ai.ts
jasonharrison: Thanks!

src/games/seabattle/ai.ts
jasonharrison: > this means that you can return an undefined move, if i gets to 10. How are you treating this case?

image

I agree that we don't need the counter.  It was sort of a hack while I was struggling to get things to work.  Will remove.

src/games/seabattle/Battle.tsx
jasonharrison: Salvos blink for 2500ms.  I added 500ms so it seems like the ai "thinks" for half a second.

src/games/seabattle/ai.ts
```diff
@@ -0,0 +1,189 @@
...
+    const absDiffLength = Math.abs(diffLength);
+    if (direction.x) {
+      for (let x = min.x - absDiffLength; x <= max.x + absDiffLength; x++) {
+        cells.push({ x: x, y: max.y });
+      }
+    } else {
+      for (let y = min.y - absDiffLength; y <= max.y + absDiffLength; y++) {
+        cells.push({ x: max.x, y: y });
+      }
+    }
```
flamecoals: You definitely dont need two for loops. You can start with the min and add both x and y of direction to the min in each iteration, so you only have one for loop.

flamecoals:
This code looks so much better! Just one thing to make it less redundant.

Tip: whenever you see an if doing separate things by direction, you can generally use vector arithmetic instead :).

PR #164, opened 2019-03-11
Add easy AI to TicTacToe

jasonharrison:
Add easy AI to TicTacToe
---
The current AI for TicTacToe uses BGIO's MCTS, and it does not let users win.

This PR adds an easy "AI" which makes moves randomly. 

The easy AI is now the default, and the MCTS AI remains as the "hard" option.

A dropdown for selecting the difficulty was also added as IGameModeExtraInfoDropdown.

### DIFF
--- a/src/App/Game/GameModePicker.tsx
+++ b/src/App/Game/GameModePicker.tsx
@@ -7,6 +7,8 @@ import CardActions from '@material-ui/core/CardActions';
 import CardContent from '@material-ui/core/CardContent';
 import CardHeader from '@material-ui/core/CardHeader';
 import Slider from '@material-ui/lab/Slider';
+import MenuItem from '@material-ui/core/MenuItem';
+import MenuList from '@material-ui/core/MenuList';
 import Avatar from '@material-ui/core/Avatar';
 import Typography from '@material-ui/core/Typography';
 import Button from '@material-ui/core/Button';
@@ -39,6 +41,11 @@ export interface IGameModeExtraInfoSlider extends IGameModeExtraInfo {
   max: number;
 }
 
+export interface IGameModeExtraInfoDropdown extends IGameModeExtraInfo {
+  type: 'dropdown';
+  options: string[];
+}
+
 export enum GameMode {
   AI = 'AI',
   OnlineFriend = 'online',
@@ -51,7 +58,8 @@ export class GameModePicker extends React.Component<IGameModePickerProps, IGameM
     return React.createElement(Link, { ...props, to }, props.children);
   }
 
-  _handleExtraInfoChange = (mode: GameMode) => (event: any, value: number) => {
+  _handleExtraInfoChange = (mode: GameMode, argValue?: any) => (event: any, value?: number) => {
+    value = value || argValue;
     const newState: IGameModePickerState = {
       ...this.state,
       extraInfo: {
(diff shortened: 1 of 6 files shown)


src/App/Game/GameModePicker.tsx
```diff
@@ -144,6 +152,34 @@ export class GameModePicker extends React.Component<IGameModePickerProps, IGameM
             />
           </div>
         );
+      } else if (extraInfo.type === 'dropdown') {
+        const dropdown = (extraInfo as IGameModeExtraInfoDropdown);
+        const list: JSX.Element[] = [];
+        dropdown.options.map((option, idx) => {
```
flamecoals: Map converts each element of a list into another element of a new list. So you should not use a side effect such push inside map.

Do list = dropdown.options.map yada yada yada return <MenuItem

src/games/tictactoe/ai.ts
```diff
@@ -6,19 +6,45 @@ interface IPlayState {
   ctx: any;
 }
 
+class TictactoeRandomBot {
+  async play(state: IPlayState, playerID: string) {
+    const cell = this.generateMove(state);
+    return this.makeMove(cell);
+  }
+  generateMove(state: IPlayState) {
```
flamecoals: Would be good to say it is random in the name

src/App/Game/GameModePicker.tsx
jasonharrison: Good idea, thanks!

src/games/tictactoe/ai.ts
jasonharrison: I actually had that and then I decided to name the class "TictactoeRandomBot".  But I agree, I'll rename the method.

src/App/Game/GameModePicker.tsx
```diff
@@ -51,7 +58,8 @@ export class GameModePicker extends React.Component<IGameModePickerProps, IGameM
     return React.createElement(Link, { ...props, to }, props.children);
   }
 
-  _handleExtraInfoChange = (mode: GameMode) => (event: any, value: number) => {
+  _handleExtraInfoChange = (mode: GameMode, argValue?: any) => (event: any, value?: number) => {
+    value = value || argValue;
```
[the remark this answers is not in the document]
jasonharrison: So the ```<Slider>``` has **onChange**={this._handleExtraInfoChange...}, but my ```<MenuItem>``` has **onClick**={this._handleExtraInfoChange...}.

I found that onClick does not pass a value to _handleExtraInfoChange:

```typescript
TS2322: Type '(event: any, value: number) => void' is not assignable to type '(event: MouseEvent<HTMLElement, MouseEvent>) => void'.
```

I got around this by passing my value as an argument (**argValue**).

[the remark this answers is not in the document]
flamecoals: this is also the part that is decreasing the coverage.
I would say, leave _handleExtraInfoChange, maybe rename it to _handleSliderChange, and create a new one _handleClickSelection, and make sure to test a click in this.

src/App/Game/GameModePicker.tsx
```diff
@@ -144,6 +152,33 @@ export class GameModePicker extends React.Component<IGameModePickerProps, IGameM
...
+            <MenuList
+              style={{
+                paddingRight: '20%',
+                display: 'flex',
+                listStyle: 'none',
+              }}
+            >
+              {list}
+            </MenuList>
+          </div>
```
flamecoals: this function is too big, can you break it down into smaller ones ? ie getDropDown and getSlider ?

PR #166, opened 2019-03-14
Dynamically generate robots.txt

jasonharrison:
Dynamically generate robots.txt


src/server.tsx
```diff
@@ -104,4 +102,22 @@ const startServer = async () => {
   });
 };
 
+function getOpenRobotsTxt() {
+  const robotstxt = ['User-agent: *'];
```
flamecoals: please use consistent variable naming. In this case, robotTxt

src/server.tsx
```diff
@@ -104,4 +102,22 @@ const startServer = async () => {
...
+  const AI = GameMode.AI;
+  const OnlineFriend = GameMode.OnlineFriend;
+  GAMES_LIST.forEach((game) => {
+    game.modes.forEach((modeInfo) => {
+      if (modeInfo.mode === GameMode.AI) {
+        robotstxt.push(`Disallow: /g/${game.code}/ai/`);
```
flamecoals: isnt AI supposed to be capital letters here?

src/server.tsx
```diff
@@ -104,4 +102,22 @@ const startServer = async () => {
...
+      }
+      if (modeInfo.mode === GameMode.OnlineFriend) {
+        robotstxt.push(`Disallow: /g/${game.code}/online/`);
```
flamecoals: arent we supposed to have a * in the end?

src/server.tsx
jasonharrison: Yes, you're right -- and robots.txt **is** case sensitive!  Thank you for catching!

src/server.tsx
jasonharrison: Apparently some robots.txt parsers do not support wildcard syntax.  /online/ and such should work and disallow all subdirectories.

PR #167, opened 2019-03-14
Check if state.ai is undefined, and if so, load before using it

### LINKED ISSUE #152: Bug - White page instead of Chess AI
This bug happens if you select a "Local friend" game mode, click on the FBG logo, and then go to "Chess AI" game mode.

jasonharrison:
This PR will fix #152

### DIFF
--- a/src/App/Game/Game.tsx
+++ b/src/App/Game/Game.tsx
@@ -110,7 +110,13 @@ export default class Game extends React.Component<IGameProps, {}> {
         clientConfig.enhancer = state.config.enhancer;
       }
       if (this.loadAI) {
-        clientConfig.ai = state.ai.bgioAI(aiLevel);
+        if (typeof state.ai === 'undefined') {
+          this.load().then(() => {
+            clientConfig.ai = state.ai.bgioAI(aiLevel);
+          });
+        } else {
+          clientConfig.ai = state.ai.bgioAI(aiLevel);
+        }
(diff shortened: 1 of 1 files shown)


src/App/Game/Game.tsx
```diff
@@ -110,7 +110,13 @@ export default class Game extends React.Component<IGameProps, {}> {
         clientConfig.enhancer = state.config.enhancer;
       }
       if (this.loadAI) {
-        clientConfig.ai = state.ai.bgioAI(aiLevel);
+        if (typeof state.ai === 'undefined') {
+          this.load().then(() => {
```
flamecoals: I think this is not addressing the root cause.... The idea is that it would only come up here AFTER the load, given that we have the "loading" flag being false

flamecoals: The solution should be:

- Revert these changes you made
- Call this.clear() first thing in the constructor OR load() (I think load() will work while constructor might not)

jasonharrison: I tried this and tests broke: 
https://semaphoreci.com/freeboardgame/freeboardgame-org/branches/issue-152/builds/2

Not sure whether things actually broke, or whether the tests need to be changed.

jasonharrison:
https://github.com/freeboardgame/FreeBoardGame.org/pull/171

PR #168, opened 2019-03-21
Give each ship a unique ID

jasonharrison:
Give each ship a unique ID (to be used in a future seabattle AI PR).

### DIFF
--- a/src/games/seabattle/game.test.tsx
+++ b/src/games/seabattle/game.test.tsx
@@ -39,10 +39,10 @@ describe('Seabattle', () => {
     const client = Client({
       game: SeabattleGame,
     });
-    client.moves.setShips(VALID_SETUP_FIRST_PLAYER);
-    const state0 = client.store.getState();
-    state0.G.ships[0].id = 2;
-    expect(validateShips(state0.G.ships).valid).toBeFalsy();
+    expect(() => {
+      client.moves.setShips([{ player: 0, id: 1, cells: [{ x: 0, y: 9 }, { x: 1, y: 9 }] },
+      { player: 0, id: 1, cells: [{ x: 2, y: 9 }, { x: 3, y: 9 }] }]);
+    }).toThrow();
   });
 
   it('should only allow correct player to set ships', () => {
@@ -135,7 +135,7 @@ describe('Seabattle', () => {
     expect(store.getState().ctx.gameover.winner).toEqual('1');
   });
 
-  it('should show victory correctly - player 11 loss', () => {
+  it('should show victory correctly - player 1 loss', () => {
     const client = Client({
       game: SeabattleGame,
     });
--- a/src/games/seabattle/game.test.tsx
+++ b/src/games/seabattle/game.test.tsx
@@ -1,5 +1,5 @@
 import { Client } from '@freeboardgame.org/boardgame.io/client';
-import { SeabattleGame, IShip, generateRandomShips, playerView } from './game';
+import { SeabattleGame, IShip, generateRandomShips, validateShips, playerView } from './game';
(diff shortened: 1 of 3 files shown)


src/games/seabattle/game.ts
```diff
@@ -244,6 +251,17 @@ export function getCellVector(a: ICell, b: ICell): ICell {
   return { x: a.x - b.x, y: a.y - b.y };
 }
 
+function validateShipsHaveUniqueIDs(ships: IShip[]): IShipsValidationResult {
+  const usedIDs: number[] = [];
+  for (const ship of ships) {
+    if (usedIDs.includes(ship.id)) {
```
flamecoals: This is O(n^2) while you could easily do it using a hashmap instead of a list and get it to be O(n)

src/games/seabattle/game.ts
```diff
@@ -53,6 +54,9 @@ export const SeabattleGame = Game({
   moves: {
     setShips(G: ISeabattleState, ctx: ICtx, ships: IShip[]) {
       const player = parseInt(ctx.playerID, 10);
+      for (let i = 0; i < ships.length; i++) {
+        ships[i] = { ...ships[i], id: i + G.ships.length };
```
flamecoals: Are you overriding the random Id you generated in the random ship functions? Are you sure this works because this is relying on G which will be different for each player?

src/games/seabattle/game.test.tsx
```diff
@@ -35,6 +35,16 @@ describe('Seabattle', () => {
     }).toThrow();
   });
 
+  it('should only allow unique IDs', () => {
+    const client = Client({
+      game: SeabattleGame,
+    });
+    client.moves.setShips(VALID_SETUP_FIRST_PLAYER);
```
flamecoals: Don't you need to update this VALID_SETUP_FIRST_PLAYER with random IDs?

src/games/seabattle/game.ts
jasonharrison: fixed by using shortids

src/games/seabattle/game.test.tsx
jasonharrison: Yes, I forgot to commit the mock.ts -- fixed

src/games/seabattle/game.ts
```diff
@@ -116,12 +118,14 @@ export const SeabattleGame = Game({
...
   let result: IShip[];
+  let shipID;
   do {
     result = [];
     for (const shipSize of VALID_SHIPS_SIZES) {
       const count: number = VALID_SHIPS_COUNT[shipSize];
       for (let i = 0; i < count; i++) {
-        result.push(randomlyGetShip(player, shipSize));
+        shipID = shortid.generate();
+        result.push(randomlyGetShip(player, shipSize, shipID));
```
flamecoals: We should avoid setShips being non deterministic given that we might need to replay the game. Please just check that they are unique and set to whatever is provided. Generate the random strings on the UI code.

src/games/seabattle/mocks.ts
```diff
@@ -11,6 +12,7 @@ export const VALID_SETUP_FIRST_PLAYER: IShip[] = [
       { x: 4, y: 0 },
     ],
     sunk: true,
+    id: shortid.generate(),
```
flamecoals: These are mocks for testing, so use deterministic IDs like "foo" and "bar" instead of random ids

src/games/seabattle/game.ts
jasonharrison: discussed on gho

src/games/seabattle/mocks.ts
jasonharrison: Good idea

PR #181, opened 2019-03-31
Chess: when playing an AI game, show you won/you lost at gameover

jasonharrison:
Chess: when playing an AI game, show you won/you lost at gameover

### DIFF
--- a/src/games/chess/board.tsx
+++ b/src/games/chess/board.tsx
@@ -170,9 +170,9 @@ export class Board extends React.Component<IBoardProps, {}> {
   }
 
   _getGameOver() {
-    // Online Multiplayer
-    if (this.props.gameArgs &&
-      (this.props.gameArgs.mode === GameMode.OnlineFriend || this.props.gameArgs.mode === GameMode.AI)) {
+    const gameArgs = this.props.gameArgs;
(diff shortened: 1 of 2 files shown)


src/games/chess/board.tsx
```diff
@@ -171,7 +171,8 @@ export class Board extends React.Component<IBoardProps, {}> {
 
   _getGameOver() {
     // Online Multiplayer
-    if (this.props.gameArgs && this.props.gameArgs.mode === GameMode.OnlineFriend) {
+    if (this.props.gameArgs &&
+      (this.props.gameArgs.mode === GameMode.OnlineFriend || this.props.gameArgs.mode === GameMode.AI)) {
```
flamecoals: add a line before the if:
```
const mode = this.props.gameArgs.mode
````
and then replace with (mode === GameMode.OnlineFriend || mode === GameMode.AI)
this will be much more legible

jasonharrison: > add a line before the if:
> ```const mode = this.props.gameArgs.mode```

This alone won't work because gameArgs could be undefined.  Yay for ternary operators!  I'm liking TS more and more.

PR #182, opened 2019-03-31
Chess: Highlight squares for the last move

### LINKED ISSUE #55: Chess: Highlight squares for the last move.

jasonharrison:
Chess: Highlight squares for the last move (closes #55)

@flamecoals, this works in multiplayer mode but does not work well with AI.  I suspect that async is the problem, can you help?  Also, what color should we make the highlighted square?

image

### DIFF
--- a/src/games/chess/board.tsx
+++ b/src/games/chess/board.tsx
@@ -126,6 +126,15 @@ export class Board extends React.Component<IBoardProps, {}> {
     for (const move of this._getMoves()) {
       result[move.to] = MOVABLE_COLOR;
     }
+    const pgnArray = this.props.G.pgn.split(' ');
+    if (pgnArray.length > 0) {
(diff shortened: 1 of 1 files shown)


src/games/chess/board.tsx
```diff
@@ -126,6 +126,15 @@ export class Board extends React.Component<IBoardProps, {}> {
     for (const move of this._getMoves()) {
       result[move.to] = MOVABLE_COLOR;
     }
+    const pgnArray = this.props.G.pgn.split(' ');
```
flamecoals: wow no, what you are trinng to do is extremely buggy. We have a Chess() object that you can use the Chess.js API and look the last move using the documented API instead. See here:

https://github.com/jhlywa/chess.js/blob/master/README.md
see history() specially with verbose: true

jasonharrison: That is much better.  Thanks!

PR #183, opened 2019-03-31
Add analytics to GameOver

jasonharrison:
Add analytics to GameOver

### DIFF
--- a/src/App/Game/GameOver.tsx
+++ b/src/App/Game/GameOver.tsx
@@ -39,12 +39,12 @@ export class GameOver extends React.Component<IGameOverProps, {}> {
       ReactGA.event({
         category: 'GameOver',
         label: this.props.gameArgs.gameCode,
-        action: this.props.result,
+        action: this.props.result,  // 'red won'
       });
     } else {
       ReactGA.event({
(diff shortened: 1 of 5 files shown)


src/App/Game/GameOver.tsx
```diff
@@ -35,6 +35,18 @@ export class GameOver extends React.Component<IGameOverProps, {}> {
     } else {
       playAgainText = 'Do you want to play again? Check out our games below.';
     }
+    if (typeof this.props.gameArgs !== 'undefined') {
+      ReactGA.event({
+        category: 'GameOver',
+        label: this.props.gameArgs.gameCode,
```
flamecoals: Dont have this if. You can assume every GameOver should have a gameCode. If some tests dont have it, change the tests, do NOT add a branch in the code just because it is easier to test, specially because that state is invalid in normal conditions (not during test).

jasonharrison: > do NOT add a branch in the code just because it is easier to test

DEEBC9B1-DF3A-413B-B767-C46609CCD45D


You are right.  Fixed :)

PR #185, opened 2019-04-01
Seabattle: Add other player screen on game end.

### LINKED ISSUE #115: Seabattle: Add other player screen on game end.
Probably a good idea to make the GameOver component receive an arbitrary component.

jasonharrison:
Seabattle: Add other player screen on game end.

Will close #115


src/App/Game/GameOver.tsx
```diff
@@ -8,10 +8,13 @@ import { Link } from 'react-router-dom';
 import Button from '@material-ui/core/Button';
 import ReplayIcon from '@material-ui/icons/Replay';
 import ReactGA from 'react-ga';
+import Card from '@material-ui/core/Card';
+import CardContent from '@material-ui/core/CardContent';
 
 export interface IGameOverProps {
   result: string;
   gameArgs?: IGameArgs;
+  otherPlayerBoard?: React.ReactNode;
```
flamecoals: I like this API (can be any node), but I think the name otherPlayerBoard is not fair, not generic enough. This could be ANY UI component that we want display at game end. I would rename it to be "extraCardContent" or something like this.

src/App/Game/GameOver.tsx
```diff
@@ -49,10 +53,24 @@ export class GameOver extends React.Component<IGameOverProps, {}> {
         <Typography variant="body1" gutterBottom={true} align="center" style={{ marginTop: '16px' }}>
           {playAgainText}
         </Typography>
+        {otherPlayerBoard}
         <GamesList />
       </FreeBoardGameBar>
     );
   }
+  _getOtherPlayerBoard = () => {
+    if (typeof this.props.otherPlayerBoard === 'undefined') {
```
flamecoals: why not only if (!this.props.otherPlayerBoard) { ?

src/App/Game/GameOver.tsx
jasonharrison: Good point, thanks!

PR #191, opened 2019-04-06
Add ReactGA event for chess moves

jasonharrison:
We already have an event for clicks, so let's add an event for a move?


jasonharrison:
@flamecoals said he would do an enhancer that will do this for all the games; closing this PR.

PR #197, opened 2019-04-10
Use meta noindex tag instead of robots.txt

### LINKED ISSUE #192: Use meta noindex tag
Use meta noindex tag instead of robots.txt

https://support.google.com/webmasters/answer/93710?hl=en

jasonharrison:
Use meta noindex tag instead of robots.txt

https://support.google.com/webmasters/answer/93710?hl=en

If running on freeboardgame.org, return 404 for /robots.txt, else, return a restrictive robots.txt

Closes #192

### DIFF
--- a/src/server.tsx
+++ b/src/server.tsx
@@ -36,16 +36,16 @@ function renderHtml(layout: string, metadata: IPageMetadata, reactHtml: string) 
 
   if (metadata.description) {
     result = result.replace(
-      '    <meta name="description" content="">\n',
-      `    <meta name="description" content="${metadata.description}">\n`);
+      '<meta name="description" content="">\n',
+      `<meta name="description" content="${metadata.description}">\n`);
   } else {
     result = result.replace(
-      '    <meta name="description" content="">\n',
+      '<meta name="description" content="">\n',
       '');
   }
 
   if (!metadata.noindex) {
-    result = result.replace('   <meta name="robots" content="noindex">\n',
+    result = result.replace('<meta name="robots" content="noindex">\n',
       '');
   }
 
--- a/src/server.tsx
+++ b/src/server.tsx
@@ -12,7 +12,7 @@ import serialize from 'serialize-javascript';
 import { StaticRouter } from 'react-router-dom';
 import { GAMES_LIST } from './games';
 import { GameMode } from './App/Game/GameModePicker';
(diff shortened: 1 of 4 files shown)


src/server.tsx
```diff
@@ -27,16 +27,28 @@ const DEV = !PROD;
...
-  result = result.replace('<title>FreeBoardGame.org</title>', `<title>${title}</title>`);
-  result = result.replace(
-    '<meta name="Description" content="">',
-    `<meta name="Description" content="${description}">`);
+
+  result = result.replace('<title>FreeBoardGame.org</title>', `<title>${metadata.title}</title>`);
+
+  if (metadata.description) {
+    result = result.replace(
+      '    <meta name="description" content="">\n',
```
flamecoals: can we remove the spaces inside the string being replaced? I think this is a very easy source of failure.

src/server.tsx
```diff
@@ -79,9 +91,7 @@ const startServer = async () => {
   server.app.use(router.allowedMethods());
   server.app.use(async (ctx: any, next: any) => {
```
flamecoals: This block can be removed altogether, no? We dont rely on robots.txt anymore?

src/server.tsx
```diff
@@ -79,9 +91,7 @@ const startServer = async () => {
...
     if (ctx.request.path === '/robots.txt') {
-      if (ctx.request.hostname.toLowerCase() === 'freeboardgame.org') {
-        ctx.response.body = OPEN_ROBOTS_TXT;
-      } else {
+      if (ctx.request.hostname.toLowerCase() !== 'freeboardgame.org') {
         ctx.response.body = RESTRICTIVE_ROBOTS_TXT;
```
flamecoals: We can probably remove the RESTRICTIVE and OPEN_ROBOTS_TXT constant too

src/server.tsx
jasonharrison: Removed, it was so the HTML was aligned

src/server.tsx
jasonharrison: OPEN_ROBOTS_TXT is not present.

RESTRICTIVE_ROBOTS_TXT is served when the Host header is not freeboardgame.org, which I thought was desired behavior.

PR #212, opened 2019-05-11
WIP i18n

jasonharrison:
Working on i18n


src/Home/Header.test.tsx
```diff
@@ -6,9 +6,9 @@ import { expect } from 'chai';
 
 describe('Header', () => {
 
-  it('should contain Chess option', () => {
+  it('should contain header', () => {
     const wrapper = mount(<MemoryRouter><Header /></MemoryRouter>);
-    expect(wrapper.text()).to.include('Free as in freedom');
+    expect(wrapper.text()).to.include('headerTitle');
```
flamecoals: I think we should find a way to make the replacement here... Otherwise all the tests are going to be hard to read. Does it work if you use <I18nextProvider></I18nextProvider> around the component ?

flamecoals: https://react.i18next.com/misc/testing#example-test-using-this-configuration this is helpful

PR #219, opened 2019-05-17
Separate bgio server and FBG

jasonharrison:
Separate bgio server and FBG (allows for updates of FBG without having to stop the bgio server)


Dockerfile
```diff
@@ -21,5 +23,6 @@ RUN yarn install
 RUN yarn run hexo generate
 
 WORKDIR /appdata
+
 # start server
```
flamecoals: Why do you have two docker files?

package.json
```diff
@@ -15,7 +15,9 @@
     "prod:build": "run-s server:build webpack:build",
     "precommit": "yarn run verify-lint && yarn run verify-tsfmt && yarn run server:build && yarn run test",
     "pre": "yarn run precommit",
-    "server": "node server-build/server.js",
+    "server:bgio": "node server-build/server_bgio.js",
```
flamecoals: I think you can leave "server" executing both

src/App/Game/Game.tsx
```diff
@@ -123,7 +123,9 @@ export default class Game extends React.Component<IGameProps, {}> {
         clientConfig.ai = state.ai.bgioAI(aiLevel);
       }
       if (this.mode === GameMode.OnlineFriend) {
-        clientConfig.multiplayer = true;
+        const server = typeof window !== 'undefined' ?
+          process.env.BGIO_SERVER_URL || `${window.location.hostname}:8001` : 'localhost:8001';
```
flamecoals: When would the window be undefined?

webpack.dev.js
```diff
@@ -12,10 +13,13 @@ module.exports = merge(common, {
   plugins: [
     new WebpackShellPlugin({
        onBuildEnd: {
-         scripts: ['node server-build/server.js'],
+         scripts: ['node server-build/server_fbg.js','node server-build/server_bgio.js'],
```
flamecoals: Does both execute in parallel in this case?

Dockerfile
jasonharrison: Left over by accident, removed bgio.Dockerfile

src/App/Game/Game.tsx
jasonharrison: Hrmph.  I think I ran into an issue testing with Jest when I added this, but I don't remember/can't reproduce now, so removing this.

webpack.dev.js
jasonharrison: Yes :)

src/App/Game/Game.tsx
jasonharrison: Aha.  It was for SSR.

```
ReferenceError: window is not defined
      at Game.render (webpack:///./src/App/Game/Game.tsx?:132:58)
      at /appdata/node_modules/react-tree-walker/dist/react-tree-walker.js:211:33
      at /appdata/node_modules/react-tree-walker/dist/react-tree-walker.js:141:36
```

PR #226, opened 2019-05-23
Lobby

flamecoals:


flamecoals:
Our bgio server has some issue:
```
curl -X POST http://charizard:8001/games/chess/create
```
returns 404 "Not found" when it shouldnt.

Examples work fine:
```
curl -X POST http://charizard:8000/games/tic-tac-toe/create 
                                        
{"gameID":"8714fbec-e30f-4c45-82f3-fecde22a7a94"}
```

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,12 @@
+import AddressHelper from '../AddressHelper';
+import request from 'superagent';
+
+export class LobbyService {
+  public static async newRoom(gameCode: string): Promise<string> {
+    const response = await request
+      .post(`http://${AddressHelper.getServerAddress()}/games/${gameCode}/create`)
```
jasonharrison: Putting http:// here  is problematic because ```AddressHelper.getServerAddress()``` might specify a protocol.
https://github.com/freeboardgame/FreeBoardGame.org/blob/6a8ec3966034cefb2bc15d10a4da058304e6da73/src/App/AddressHelper.ts#L5

In our production environment, ```process.env.BGIO_SERVER_URL=https://freeboardgame.org```, so this would be ```http://https://freeboardgame.org/games...```

I removed the http:// from this and changed getServerAddress() to always return a protocol.

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,43 @@
+import AddressHelper from '../AddressHelper';
+import request from 'superagent';
+
+export interface IPlayerInRoom {
+  ID: number;
```
flamecoals: id of what ? Player ID? room ID?

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,43 @@
...
+  name: string;
+  credentials?: string;
```
flamecoals: this is a single credential or more than one? why credentialS ?

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static async joinRoom(room: IRoomMetadata, player: IPlayerInRoom): Promise<string> {
+    const response = await request
+      .post(`${AddressHelper.getServerAddress()}/games/${room.gameCode}/${room.roomID}/join`)
+      .send({
+        playerID: player.ID,
+        playerName: player.name,
+      });
+    return response.body.playerCredentials;
```
flamecoals: joinRoom should be the one updating the localStorage with credentials

src/App/Lobby/NewRoom.tsx
```diff
@@ -0,0 +1,54 @@
...
+  constructor(props: INewRoomProps) {
+    super(props);
+    const gameCode = props.match.params.gameCode;
+    if (!(gameCode in GAMES_MAP)) {
+      this.state = { error: true };
+      return;
+    }
+    LobbyService.newRoom(gameCode)
+      .then((newRoom) => {
+        localStorage.setItem(newRoom.room.roomID, JSON.stringify(newRoom.initialPlayer));
```
flamecoals: no no no no NewRoom should only be the view. The localStorage management should be encapsulated inside LobbyService OR in its own Service

src/App/Lobby/LobbyService.ts
flamecoals: I think I initially coded and I assumed that getServerAddress would always return host:port. So we found a bug in the original code...

jasonharrison: Well, kinda.  BGIO would auto-insert the http:// -- it probably not the best idea for me to rely on BGIO to do this, but fixed now :)

src/App/Lobby/LobbyService.ts
jasonharrison: playerID, renamed, thanks

src/App/Lobby/LobbyService.ts
jasonharrison: One, made singular

jasonharrison: I did this because BGIO returns 'playerCredentials' -- I think I agree that credential makes more sense, though.

https://github.com/freeboardgame/FreeBoardGame.org/blob/cbb89332e46a8373c415a921cc43149302294982/src/App/Lobby/LobbyService.ts#L41

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,96 @@
...
+    return player;
+  }
+
+  public static setCredential(credential: IPlayerInRoom): void {
+    localStorage.setItem('fbgCredential', JSON.stringify(credential));
+  }
+
+  public static genPlayerID(room: IRoomMetadata): number {
+    const players = room.players;
+    const playersWithNames: any = [];
```
flamecoals: why any? The type is supposed to be IPlayerInRoom, right?

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,96 @@
...
+    for (const player of players) {
+      if (player.name) {
```
flamecoals: can a player ever have no name?

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,96 @@
...
+        playersWithNames.push(player);
+      }
+    }
+    return playersWithNames.length === 2;
+  }
+
+  public static getCredential(roomID: string): IPlayerInRoom {
```
flamecoals: Why is this called getCredential if it returns IPlayerInRoom which is much more than a credential?

src/App/Lobby/Room.tsx
```diff
@@ -0,0 +1,136 @@
...
+    const roomID = this.props.match.params.roomID;
+    this.state = {
+      nameTextBox: '',
+      room: {
+        gameCode: this.props.match.params.gameCode,
+        ready: false,
+        players: [],
+        roomID,
+      },
+      player: LobbyService.getCredential(roomID),
```
flamecoals: this is inconsistent an confusing. The same data you called PlayerInRoom, Credential, and Player.

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  public static async roomReady(room: IRoomMetadata): Promise<boolean> {
+    const roomMetadata: IRoomMetadata = await this.roomMetadata(room);
+    const players = roomMetadata.players;
+    const playersWithNames: any = [];
+    for (const player of players) {
+      if (player.name) {
+        playersWithNames.push(player);
+      }
+    }
```
flamecoals: this for loop is duplicated at genPlayerID... And it seems like you are using to count how many players. First I want to understand why you check the name (is it possible to have players without name in the first place?) and second why you dont reuse it in a private method

src/App/Lobby/LobbyService.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  public static async roomMetadata(room: IRoomMetadata): Promise<IRoomMetadata> {
+    const response = await request
+      .get(`${AddressHelper.getServerAddress()}/games/${room.gameCode}/${room.roomID}`);
+    const roomMetadata: IRoomMetadata = response.body;
+    return roomMetadata;
+  }
+
+  public static async roomReady(room: IRoomMetadata): Promise<boolean> {
```
flamecoals: returning a boolean is not a good idea, you might better off if you return the list of players or or  the whole roomMetadata so you can reuse this method more.

src/App/Lobby/Room.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export class Room extends React.Component<IRoomProps, IRoomState> {
+  private timer: any;  // fixme
+  constructor(props: IRoomProps) {
+    super(props);
+    const roomID = this.props.match.params.roomID;
+    this.state = {
+      nameTextBox: '',
+      room: {
```
flamecoals: call this roomMetadata to be consistent.

src/App/Lobby/Room.tsx
```diff
@@ -0,0 +1,136 @@
...
+        ready: false,
+        players: [],
+        roomID,
+      },
+      player: LobbyService.getCredential(roomID),
+    };
+  }
+  render() {
+    let LoadingPage;
+    const playerInRoom = this.state.room.players[this.state.player.playerID];
```
flamecoals: dont you have this.state.player? whats the difference? thats confusing

src/App/Lobby/Room.tsx
```diff
@@ -0,0 +1,136 @@
...
+          this.setState((oldState) => {
+            return { ...oldState, room };
+          });
+        }
+      }, () => {
+        throw new Error('failed to check if room ready');
+      });
+  }
+
+  _getNamePrompt = (name?: string) => {
```
flamecoals: this method seems to never be called! When are you triggering this page?

PR #242, opened 2019-05-27
Take 6! game

JosefKuchar:
Implementation of https://en.wikipedia.org/wiki/6_Nimmt!


jasonharrison:
Looks like the PR failed tests... Do tests pass on your local machine?  `yarn run pre`

Perhaps you need to merge in master?

src/games/takesix/game.ts
```diff
@@ -0,0 +1,219 @@
...
+              ...G.decks[id],
+            ],
+          },
+        }),
+        decks: Object.values({
+          ...G.decks,
+          [id]: [card],
+        }),
+      };
+    }
```
flamecoals: from line 79-96 you reuse this logic both here and from lines 42-58. I think it would be much more readable if you create a separate function "moveToHand" that given G, list of cards and player, returns the new state.

src/games/takesix/game.ts
```diff
@@ -0,0 +1,219 @@
...
+      if (G.end === true) {
+        const scoreboard = G.players.map((player, i) => ({
+          playerID: i,
+          penaltyPoints: player.penaltyCards.reduce(
+            (acc, card) => acc + card.value,
+            0,
+          ),
+        })).sort((a, b) => a.penaltyPoints - b.penaltyPoints);
+
+        if (scoreboard[0].penaltyPoints === scoreboard[1].penaltyPoints) {
```
flamecoals: you can make the draw more generic (that works for any number of players) by comparing the first element of the scoreboard with the last.

src/games/takesix/game.ts
```diff
@@ -0,0 +1,219 @@
...
+        onPhaseEnd: (G: IG) => {
+          if (G.players[0].cards.length === 0) {
+            G.end = true;
+          }
+        },
+      },
+    },
+    endGameIf: (G: IG) => {
+      if (G.end === true) {
+        const scoreboard = G.players.map((player, i) => ({
```
flamecoals: you probably going to need to show the scoreboard in the UI too, so might be useful this to live in its own function

src/games/takesix/game.ts
```diff
@@ -0,0 +1,219 @@
...
+          return { winner: scoreboard[0].playerID.toString() };
+        }
+      }
+    },
+  },
+  // playerView: PlayerView.STRIP_SECRETS,
+  setup: (ctx): IG => {
+    // Generate deck
+    const deck = ctx.random.Shuffle(
+      new Array(104).fill(0).map((_, i) => {
```
flamecoals: wouldnt it be clearer if you do:

[ 
  ...new Array(20).fill(2),
  ...new Array(10).fill(3)

and etc?

flamecoals:
Great work on this! I am really impressed with this game and your implementation, simple and extremely fun :D

src/games/takesix/game.ts
JosefKuchar: Can you explain it little bit more? I don't get it 😄

flamecoals: Your function generates 76 cards with value 1, 9 cards with value 2, 10 cards with value 3,  8 cards with value 5, and 1 card with value 7. Then, you shuffle the cards. It would be easier to understand if this was generated more directly instead of using the % function. The code would look like this:
```
const deck = ctx.random.Shuffle([ 
    ...new Array(76).fill(1), 
    ...new Array(9).fill(2), 
    ...new  Array(10).fill(3), 
    ...new  Array(8).fill(5), 
    7 
].map((value, i) => new Card(i+1, value, null));
```

flamecoals: Oh, I just noticed that my code is not the same as yours, it would have the wrong card "number". Interesting.

flamecoals: I just noticed that in the wikipedia theses cards are defined with the multiples. So your implementation makes sense. Sorry for the noise!

flamecoals:
Some stuff to do blocking the merge:
a. Replace the thumbnail image
b. Use nicknames instead of player 0, 1 etc
c. Allow 2-10 players

Other improvements that are non blocking:
- Animations
- Some sounds when an action happen

I am going to work on c) now.

flamecoals:
OK, after this PR: https://github.com/freeboardgame/FreeBoardGame.org/pull/253 is merged, it should be easy to support more than two players

JosefKuchar:
b) is done

PR #244, opened 2019-05-29
Better i18n with fallback to English

jasonharrison:
Better i18n with fallback to English

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -7,7 +7,7 @@ import About from '../About/AboutAsync';
 import getMessagePage from './MessagePage';
 import ReactGA from 'react-ga';
 import { getPageMetadata } from '../metadata';
-import { registerLang, setCurrentLocale } from 'ts-easy-i18n';
+import { registerLang, setCurrentLocale } from '@freeboardgame.org/i18n';
 import translations from './translations';
 
 ReactGA.initialize('UA-105391878-1');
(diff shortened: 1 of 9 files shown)


src/App/translations/cs.ts
```diff
@@ -1,11 +1,11 @@
 const cs = {
-    about: {
-        header: 'Informace o FreeBoardGame.org',
-    },
-    messagePage: {
-        invalidLocale: 'Neplatný jazyk',
-        goHome: 'Domů',
-    },
+  about: {
+    headeraaaaa: 'Informace o FreeBoardGame.org',
```
flamecoals: is this correct? headeraaaaa ?

src/App/App.tsx
```diff
@@ -37,6 +37,9 @@ const withI18n = (WrappedComponent: any) => {
       }
       // set language based on URL, default to English
       const lang = locale || 'en';
+      if (lang !== 'en') {
+        registerLang('en', translations.en);
```
flamecoals: so english will always be registered?
I think it would be clearer if you just wrote:
registerLang('en', translations.en);
if (lang !== 'en') {
  registerLang(lang, transaltions[lang]);
}

src/App/translations/cs.ts
jasonharrison: No... I reverted this change with the edited file still open in my IDE, and then it auto-saved the modified version.  Thanks for catching.

PR #251, opened 2019-06-01
Allow users to change their nickname

jasonharrison:
Allow users to change their nickname

~~Waiting for changes to be made to the BGIO API~~


src/App/Lobby/LobbyService.ts
```diff
@@ -47,6 +47,22 @@ export class LobbyService {
     this.setCredential(player, credential);
   }
 
+  public static async renameUser(gameCode: string, player: IPlayerInRoom, newName: string): Promise<void> {
```
flamecoals: I don't see you using this. Delete it.

src/App/Lobby/Room.tsx
```diff
@@ -47,7 +48,7 @@ export class Room extends React.Component<IRoomProps, IRoomState> {
       return <LoadingPage />;
     }
     if (!nickname) {
-      return this._getNamePrompt();
+      return <FreeBoardGameBar>{this._getNamePrompt()}</FreeBoardGameBar>;
```
flamecoals: Any reason to move the FBG bar here? Can you move back?

src/App/Lobby/LobbyService.ts
jasonharrison: https://github.com/freeboardgame/FreeBoardGame.org/blob/d46a426cf36401f5c640cb93071b0e387cc7d748/src/App/Lobby/Room.tsx#L144

jasonharrison: Oh, you meant leaveRoom.  Ok, removed.

src/App/Lobby/Room.tsx
jasonharrison: I didn't want to include the FBG bar in _getNamePrompt() because I don't always want the fbg bar.

https://github.com/freeboardgame/FreeBoardGame.org/blob/d46a426cf36401f5c640cb93071b0e387cc7d748/src/App/Lobby/Room.tsx#L65

flamecoals: If you are reusing everything on _getNamePrompt, why don't you move this to the change nickname component instead?

PR #259, opened 2019-06-05
Add more i18n

jasonharrison:
Add more i18n

### DIFF
--- a/src/App/translations/en.ts
+++ b/src/App/translations/en.ts
@@ -1,6 +1,11 @@
+/* eslint-disable prettier/prettier */
 const en = {
   about: {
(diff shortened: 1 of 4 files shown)


src/App/translations/en.ts
```diff
@@ -1,6 +1,10 @@
+/* eslint-disable prettier/prettier */
```
flamecoals: Can we disable a specific rule instead of everything?

jasonharrison: This is only disabling prettier, which is the automatic code formatter (it disliked the long line).

Other eslint rules are still on.

Issue #95, opened 2019-02-18
Warning: Material-UI: you are using the deprecated typography variants that will be removed in the next major release.


jasonharrison:
console.error node_modules/warning/warning.js:
Warning: Material-UI: you are using the deprecated typography variants that will be removed in the next major release.
 Please read the migration guide under https://material-ui.com/style/typography#migration-to-typography-v2

PR #260, opened 2019-06-05
Fix a lot of console errors/warnings related to material-ui and components 

jasonharrison:
Fix a lot of console errors/warnings related to material-ui and components 

Improving (but did not completely fix) #95

### DIFF
--- a/src/App/App.tsx
+++ b/src/App/App.tsx
@@ -12,8 +12,15 @@ import { getPageMetadata } from '../metadata';
 import { registerLang, setCurrentLocale } from '@freeboardgame.org/i18n';
 import translations from './translations';
 import SSRHelper from './Helpers/SSRHelper';
+import { MuiThemeProvider, createMuiTheme } from '@material-ui/core/styles';
 
 ReactGA.initialize('UA-105391878-1');
+
+const theme = createMuiTheme({
+  typography: {
+    useNextVariants: true,
+  },
+});
 
 const withGA = (WrappedComponent: any) => {
   class Wrapper extends React.Component<{}, {}> {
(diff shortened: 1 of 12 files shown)


src/App/App.tsx
```diff
@@ -63,21 +70,23 @@ class Main extends React.Component<{}, {}> {
...
-        <Route
-          exact={true}
-          path={`${BASEPATH}/g/:gameCode/:mode/:matchCode/:playerID`}
-          component={withWrappers(Game)}
-        />
-        <Route path={`${BASEPATH}/room/new/:gameCode/:numPlayers`} exact={true} component={withWrappers(NewRoom)} />
-        <Route path={`${BASEPATH}/room/:gameCode/:roomID`} exact={true} component={withWrappers(Room)} />
-        <Route exact={true} component={withWrappers(getMessagePage('error', 'Not Found'))} />
-      </Switch>
+      <MuiThemeProvider theme={theme}>
```
flamecoals: What's the purpose or this change? Why moving the muithemeprovider here without deleting anywhere else?

jasonharrison: https://github.com/mui-org/material-ui/issues/13175#issuecomment-428480755

I was trying to get rid of some of the warnings, and you need to have a theme to use Typography v2.  It didn't eliminate all of the warnings, though.

flamecoals: But we should have a theme everywhere we test/run the app, no? If we don't, can't we fix exactly where we don't instead of trying to apply this everywhere (and possibly twice?)

jasonharrison: Not sure what you mean by this.  AFAIK we don't have any other themes across the app?

PR #283, opened 2019-06-10
Make webpack dev server work on all hosts

flamecoals:


jasonharrison:
Approved since this was the behavior before switching to webpack-dev-server

PR #289, opened 2019-06-10
Use material ui's typography in takesix

jasonharrison:


src/games/takesix/ScoreBadges.tsx
```diff
@@ -17,9 +18,15 @@ export class ScoreBadges extends React.Component<IScoreBadgesProps, {}> {
       return (
         <div className={css.ScoreBadge} key={score.playerID}>
           <span className={css.Nickname} style={{ fontWeight: isSelf ? 'bold' : undefined }}>
-            {nickname}
+            <Typography style={{ color: 'white', display: 'inline' }} variant="body2">
```
flamecoals: In this case here, you can leverage that we have the Nickname and the ScoreBadge class already and use a selector to change these things only once in the CSS. Instead of changing twice in the styles.

PR #292, opened 2019-06-11
Add sound to chess

### LINKED ISSUE #287: Sound when making a move in checkers
Sound when making a move in checkers

I am working on this :)

jasonharrison:
Add sound to chess

Will close #287


src/games/chess/sound.ts
```diff
@@ -0,0 +1,6 @@
+import MoveSound from './media/move.mp3';
+
+export const playSound = () => {
+  const sound = new Audio(MoveSound);
```
flamecoals: Does this mean it loads the file every time it tries to play? If I am not mistaken we used to store the Audio as a global in other games?

PR #295, opened 2019-06-11
Nine men's morris game

JosefKuchar:
TODO
- [x] Check if there is valid removePiece move after creating a mill
- [x] Check if there is valid movePiece move
- [x] Local game mode


flamecoals:
Really nice game Josef! I just played it and it is really nice :D

Some todos before we merge:
- Black pieces have low contrast with the black background, we can probably change the black/white to other colors.
- Clicking on a spot or a piece sometimes is hard. We should make the circles larger when they are clickable, or add a larger, invisible square on front of it that is clickable too
- We should highlight when a piece was selected

Tests are simple and effective, nice :) I might add just one or two other cases for testing the UI directly (clicking on an element and so on)

PR #302, opened 2019-06-19
Add text instructions to Chess

jasonharrison:
Add text instructions to Chess


flamecoals:
Can you talk quickly about each piece moves?

PR #308, opened 2019-06-20
Fix flaws in logic [Nine men's morris]

JosefKuchar:
I completely overlooked you can actually remove opponent's piece when all his pieces are in mills. Also I missed that the game should end when there are no valid moves. Now it's like it should be.


src/games/ninemensmorris/game.test.ts
```diff
@@ -28,76 +28,47 @@ it('should declare player 1 as the winner', () => {
...
-  p1.moves.movePiece(5, 13);
-  p0.moves.movePiece(3, 5);
-  p1.moves.movePiece(16, 23);
-  p0.moves.movePiece(5, 3);
-  p0.moves.removePiece(13);
-  p1.moves.movePiece(23, 13);
-  p0.moves.movePiece(11, 9);
-  p1.moves.movePiece(13, 5);
+  p1.moves.movePiece(14, 13);
+  p1.moves.removePiece(9);
```
flamecoals: It seems like you are testing these corner cases of the rules here. Can you add comments for the moves that test that?

JosefKuchar: The edge case happens just one time in this scenario

PR #326, opened 2019-06-26
Cornerus

JosefKuchar:
Game similar to https://en.wikipedia.org/wiki/Blokus
TODO

- [x] Victory condition
- [x] Status bar
- [x] Scoreboard
- [x] UI tests
- [x] Description
- [x] Fix end game bugs
- [x] 3-player mode
- [x] Thumbnail
- [x] Fix `Warning: Failed prop type: Invalid prop 'children' supplied to 'Grid'`


flamecoals:
Great game again Josef :D. Some improvements I want to do before merging:
- We should give more feedback to the user when the piece isn't placed in a valid spot. Things like making the piece transparent, or changing the Check button to invalid might help
- Some buttons are more common/important than others, but everything has the same size right now. I think we could make the Check button more highlighted (with a background), seconded by the buttons that switch pieces. Similarly, we should "hide" the X button, as it isnt intended to be used often. We could even place it under a menu. (@jasonharrison  already implemented a menu option on the top bar for Seabattle)
- I think we can increase the coverage of the rules to closer to 100%. Rules generally are easy to test and provide great benefit. I can help you with this later. Right now rules are at 88% and the board at 83%, and our average for the repo is 93%

I will give more detailed feedback about the code later

Not blocking:
- For the future we might want to support gestures, would be great to be able to rotate, change pieces and etc with gestures like swap on mobile.

flamecoals:
I created a "common" folder and moved the Scoreboard to this folder. This way, we can re-use the same component across games (as we already have a scoreboard for take six and etc). I will create an issue for us to refactor to use the same common components everywhere

PR #355, opened 2019-07-02
Increase coverage for GameBoardWrapper

jasonharrison:
Increase coverage for GameBoardWrapper

### DIFF
--- a/src/App/Game/GameBoardWrapper.test.tsx
+++ b/src/App/Game/GameBoardWrapper.test.tsx
@@ -14,7 +14,7 @@ class MockBoard extends React.Component<any, any> {
 }
 
 describe('GameBoardWrapper', () => {
-  it('should show warning', () => {
+  it('should not show warning', () => {
     const Board = gameBoardWrapper({
       board: MockBoard,
       gameArgs: {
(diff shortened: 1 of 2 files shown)


src/App/Game/GameBoardWrapper.test.tsx
```diff
@@ -13,26 +13,41 @@ class MockBoard extends React.Component<any, any> {
...
+  it('should dismiss sharing', () => {
+    const Board = gameBoardWrapper({
+      board: MockBoard,
+      gameArgs: {
+        gameCode: 'chess',
+        mode: GameMode.OnlineFriend,
+      },
+    });
+    const wrapper = Enzyme.mount(<Board isConnected={true} G={{ pgn: '' }} ctx={{}} />);
+    (wrapper.instance() as any)._dismissSharing();
```
flamecoals: what does this do? We dont have this functionality in the UI. Maybe we should delete the implementation instead?

jasonharrison: You are right.  We used _dismissSharing before lobby so I didn't realize.  Removed the test and implementation.

PR #356, opened 2019-07-02
Increase coverage for LobbyService

jasonharrison:
Increase coverage for LobbyService

### DIFF
--- a/src/App/Lobby/LobbyService.test.tsx
+++ b/src/App/Lobby/LobbyService.test.tsx
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { LobbyService, IPlayerInRoom } from './LobbyService';
+import { LobbyService, IPlayerInRoom, IStoredCredentials } from './LobbyService';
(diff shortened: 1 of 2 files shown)


src/App/Lobby/LobbyService.test.tsx
```diff
@@ -47,9 +47,37 @@ describe('New Room', () => {
...
+      roomID: 'fooroom',
+      currentUser: undefined,
+      numberOfPlayers: 1,
+    });
+  });
+
+  it('should get room metadata with currentUser', async () => {
+    const mockResponse = { body: { players: [{ id: 0, name: 'Jason', roomID: 'fooroom' }] } };
+    request.get = jest.fn().mockReturnValue(mockResponse);
+    Storage.prototype.getItem = () => JSON.stringify({ fooroom: { playerID: 0, credential: 'foocredential' } });
```
flamecoals: can you cast the type of the argument of the stringify to be the type we already defined as being the stored credentials?

PR #363, opened 2019-07-04
Increase coverage for Game.tsx

jasonharrison:
Increase coverage for Game.tsx


jasonharrison:
I removed 

https://github.com/freeboardgame/FreeBoardGame.org/pull/363/files#diff-09caaf2cc4341740e4180d6648a9159dL175 

I sort of remember talking about this with @flamecoals before, I don't think this throw is reachable.

Issue #366, opened 2019-07-04
Refactor: re-use common components across games


flamecoals:
Multiple games now have their own version of Scoreboard:

https://github.com/freeboardgame/FreeBoardGame.org/blob/master/src/games/reversi/Scoreboard.tsx
https://github.com/freeboardgame/FreeBoardGame.org/blob/master/src/games/takesix/Scoreboard.tsx

 and some have their own version of ScoreBadges:

https://github.com/freeboardgame/FreeBoardGame.org/blob/master/src/games/takesix/ScoreBadges.tsx
https://github.com/freeboardgame/FreeBoardGame.org/blob/master/src/games/reversi/ScoreBadges.tsx

We should avoid this duplication of code by creating common components in src/common.

PR #371, opened 2019-07-06
Common components

JosefKuchar:
#366


flamecoals:
:D Great. this looks much better :). It will make it easier for future games to implement these features

PR #373, opened 2019-07-06
Cornerus: check if out of bounds when dragging

jasonharrison:
Cornerus: check if out of bounds when dragging


src/games/cornerus/board.tsx
```diff
@@ -152,11 +152,13 @@ export class Board extends React.Component<IBoardProps, IBoardState> {
     const x = Math.round(coords.x);
     const y = Math.round(coords.y);
     const transform = { ...this.state.piece.transform, x, y };
-    this.setState({
-      ...this.state,
-      piece: { ...this.state.piece, transform },
-      validTransform: this.isTransformValid(this.state.piece.data, transform),
-    });
+    if (inBounds(x, y)) {
```
JosefKuchar: You can't just check if transform is out of bounds. There are cases where transform is out of bounds, but pieces are perfectly fine. Or from the opposide perspective: The transform is in bounds but the pieces are not. Solution: Reuse the bound check from getPositions (in game.ts)

PR #387, opened 2019-07-10
Reduce difficulty playing against Chess AI at lower levels

jasonharrison:
Reduce difficulty playing against Chess AI at lower levels

This seemed to make my Chess AI games last longer, but I still found it difficult to win even at level 1.


flamecoals:
I dont think that's a good solution for the problem, I think we should have a separate engine for the easier levels. I might work on it later

PR #388, opened 2019-07-10
Upgrade redux to 4.0.4

jasonharrison:


jasonharrison:
> Do we use redux at all in fbg? If we don't, we could remove this dependency

Yes:

https://github.com/freeboardgame/FreeBoardGame.org/blob/31bc737fca194e78bc2ed1dbefdd3fff7b16d9a2/src/App/Game/Game.tsx#L8

PR #397, opened 2019-07-16
Increase test coverage in LobbyService.ts

jasonharrison:
Increase test coverage in LobbyService.ts

### DIFF
--- a/src/App/Lobby/LobbyService.ts
+++ b/src/App/Lobby/LobbyService.ts
@@ -96,7 +96,9 @@ export class LobbyService {
   public static getCredential(roomID: string): IPlayerCredential | undefined {
     // return an empty IPlayerInRoom object if the player's identity is for another room
     const credentials: IStoredCredentials = JSON.parse(localStorage.getItem(FBG_CREDENTIALS_KEY));
-    return credentials[roomID];
+    if (credentials) {
+      return credentials[roomID];
+    }
   }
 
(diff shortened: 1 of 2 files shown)


src/App/Lobby/LobbyService.ts
```diff
@@ -96,9 +96,7 @@ export class LobbyService {
   public static getCredential(roomID: string): IPlayerCredential | undefined {
     // return an empty IPlayerInRoom object if the player's identity is for another room
     const credentials: IStoredCredentials = JSON.parse(localStorage.getItem(FBG_CREDENTIALS_KEY));
-    if (credentials) {
-      return credentials[roomID];
-    }
+    return credentials[roomID];
```
flamecoals: Isn't this going to break if localStorage is empty?

jasonharrison: I thought so too, but nonexistent keys are undefined.

```javascript
var x = {foo: 'bar'}
    undefined

x['foo']
    "bar"

x['bar']
    undefined

```

flamecoals: But you are assuming that JSON.parse will return an object. It might return null/undefined instead, and then you get a error trying to access a key of null?

jasonharrison: I didn't think of that.  You are right.  👍

PR #404, opened 2019-07-18
Routing: only allow valid gamemodes (ai, local)

jasonharrison:
Routing: only allow valid gamemodes (ai, local)

@flamecoals Is there a better way to do what I did here?  
 https://github.com/freeboardgame/FreeBoardGame.org/blob/d787cd872ef7d35f147a21a659d88f70b41a3cda/src/App/Game/Game.tsx#L111-L115

### DIFF
--- a/src/App/Game/Game.tsx
+++ b/src/App/Game/Game.tsx
@@ -107,6 +107,14 @@ export default class Game extends React.Component<IGameProps, IGameState> {
     if (!this.gameDef) {
       return <MessagePageClass type={'error'} message={'Game Not Found'} />;
     }
+    const validGameModes: string[] = [];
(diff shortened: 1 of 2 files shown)


src/App/Game/Game.tsx
```diff
@@ -107,6 +107,14 @@ export default class Game extends React.Component<IGameProps, IGameState> {
     if (!this.gameDef) {
       return <MessagePageClass type={'error'} message={'Game Not Found'} />;
     }
+    const validGameModes: string[] = [];
```
flamecoals: Replace this with an one liner:

const valisGameModes = this.gameDef.modes.map((m) => m.toLowerCase())

PR #413, opened 2019-07-30
Fix husky running tests twice, upgrade various deps

jasonharrison:
Fix husky running tests twice, upgrade various deps


flamecoals:
Thanks!

PR #419, opened 2020-01-01
Four in a row

gk-patel:
Hi Guys, 

I have added Connect Four to the list of Games. 

Honestly speaking, I did my best with the CSS (ie. look-and-feel), but I am no CSS expert. If you want to adjust the colors of the game, please feel free to do it. In that case, you will need to make changes in two places, 1) Shape.tsx and 2) the varaible "localPlayerNames" in constanst.ts to adjust the names of the chips to the new colour you might choose. 

Thanks and a happy new year.


jasonharrison:
Hi @gk-patel !  Thanks for the PR. I think you're aware of this since you're in our Discord, but I'm in the midst of some significant changes to the website logic code. 

I'll try to check out your code soon, and I might make some minor changes to it so it works on the new site.

jasonharrison:
Hi @gk-patel , in addition to the comments I pointed out in the review, I'm wondering whether there are better colors to use other than green/blue.  Maybe red/green?

gk-patel:
Hi, Thanks for your feedback,

Sure I am open to changing the colours, as I said earlier, I am no CSS expert. And FYI, if you change colour you will only need to change the colour code and the name of the colour which is stored in a variable. Thereafter everything else will be adapted automatically.

JosefKuchar:
I would prefer material UI defined colors to stay consistent

PR #436, opened 2020-01-17
Add bgio:dev script

jasonharrison:
Add bgio:dev script for auto-recompile/restart of BGIO server on changes


package.json
```diff
@@ -7,6 +7,7 @@
     "pre": "run-s lint format:check build:bgio test:coverage test:cypress report:combined",
     "autopre": "run-s format:staged lint build:bgio test:coverage test:cypress report:combined",
     "dev": "node server.js",
+    "bgio:dev": "nodemon -e js,jsx,ts,tsx --exec yarn run bgio",
```
JosefKuchar: Hmm, Do we need to restart the bgio server when jsx/tsx file changes?

jasonharrison: After thinking about it, I'm not sure.  I changed this.

PR #462, opened 2020-04-04
WIP Lobby / Matchmaking

jasonharrison:
Lobby / matchmaking


backend/src/User.ts
```diff
@@ -0,0 +1,66 @@
...
+  UNKNOWN_EMAIL,
+  BAD_PASSWORD,
+}
+
+export enum REG_RESULT_CODE {
+  SUCCESS,
+  NON_UNIQ_EMAIL,
+  NON_UNIQ_USERNAME,
+  WEAK_PASSWORD,
+}
```
flamecoals, line 15: Please make sure this is shared between the frontend and the backend.

backend/src/User.ts
```diff
@@ -0,0 +1,66 @@
+import { UserDBEntity } from './entities/UserDBEntity';
+import * as bcrypt from 'bcrypt';
+
+export enum AUTH_RESULT_CODE {
```
flamecoals, line 4: all capslocks are reserved for constants. This isnt a constant, it is a enum (which is more like a class/type). Use camel case instead.

backend/src/entities/UserDBEntity.ts
```diff
@@ -0,0 +1,16 @@
+import { Entity, Column, BaseEntity, PrimaryColumn } from 'typeorm';
+
+@Entity()
+export class UserDBEntity extends BaseEntity {
```
flamecoals, line 4: UserDbEntity?

backend/src/server.ts
```diff
@@ -0,0 +1,81 @@
...
+      return;
+    }
+    const user = new User(username);
+    const addResult = await user.addUser(email, password);
+    const result: RegResponse = { status: addResult };
+    res.send(result);
+  });
+
+  app.post('/api/users/auth', csrfProtection, async (req, res) => {
+    console.log('Cookies: ', req.cookies);
```
flamecoals: remove

backend/src/User.ts
```diff
@@ -0,0 +1,66 @@
...
+    const passwordHash = user.passwordHash;
+    const isCorrectPassword = bcrypt.compareSync(password, passwordHash);
+    if (!isCorrectPassword) {
+      return AUTH_RESULT_CODE.BAD_PASSWORD;
+    }
+    return AUTH_RESULT_CODE.SUCCESS;
+  }
+
+  /** Adds a user to the database with the given password. */
+  async addUser(rawEmail: string, password: string) {
```
flamecoals, line 45: `save()` instead?

backend/src/User.ts
```diff
@@ -0,0 +1,66 @@
...
+export enum REG_RESULT_CODE {
+  SUCCESS,
+  NON_UNIQ_EMAIL,
+  NON_UNIQ_USERNAME,
+  WEAK_PASSWORD,
+}
+
+const SALT_ROUNDS = 10;
+
+export default class User {
```
flamecoals, line 19: This class is weird to me... What does it represent, a user? But then you only store the username, nothing else... So it is more like the username ? with some database functions?

Honestly, I think a "User" should be something like this:

```
interface User {
  username: string;
  password?: string;
  email?: string;
}
```

And these methods are only "helper functions" that use "User"

components/App/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,307 @@
...
+                onClick={this._submitButtonClicked}
+                style={{ marginTop: '16px' }}
+              >
+                {actionText}
+              </Button>
+            </div>
+          </CardContent>
+        </Card>
+      </AlertLayer>
+    );
```
flamecoals, line 216: This is YUUUGE and hard to follow, can you break it down into submethods?

components/App/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,307 @@
...
+class LoginForm extends React.Component<Props, State> {
+  emailInput: HTMLDivElement;
+  constructor(props) {
+    super(props);
+    this.state = {
+      username: '',
+      email: '',
+      password: '',
+      submitButtonEnabled: false,
+      isRegistering: false,
```
flamecoals, line 56: This is probably a bad idea. The UI looks very different from when the user is logging in and when they are registering.

It is better to have two classes, one for each case. And then the components that are re-used across them should be their own React components.

components/App/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,307 @@
...
+                  InputProps={{
+                    startAdornment: (
+                      <InputAdornment position="start">
+                        <VpnKeyIcon />
+                      </InputAdornment>
+                    ),
+                  }}
+                />
+              </div>
+              {isRegistering && (
```
flamecoals, line 167: isRegistering or not makes it really hard to follow.

components/App/Auth/LoginForm.tsx
```diff
@@ -0,0 +1,307 @@
...
+            title={
+              <Typography style={{ marginLeft: '25%' }} variant="h6" component="h3" noWrap={true}>
+                {actionText}
+              </Typography>
+            }
+          />
+          <CardContent>
+            <FormControl className={classes.margin}>
+              <div>
+                <TextField
```
flamecoals, line 105: each textfield could be its own component/method.

pages/_app.tsx
```diff
@@ -44,6 +44,7 @@ class defaultApp extends App {
     );
   }
   static async getInitialProps({ Component, ctx }) {
+    // console.log((ctx.req as any).SSROnlyUtil.secret());
```
flamecoals, line 47: Remove

server/web.ts
```diff
@@ -69,6 +75,19 @@ app
   .then(() => {
     const server = express();
     server.disable('x-powered-by');
+    server.use(cookieParser());
+
+    // :method :url :status :response-time ms - :res[content-length]
```
flamecoals, line 80: Remove commented code

pages/_app.tsx
jasonharrison: Thanks :)

PR #481, opened 2020-04-09
Add Secret Codes

JvSomeren:
You might want to rename the game because of licensing.
I haven't added any tests but other than that everything should work pretty decent.


flamecoals:
Thank you for your PR, I will try to review it tonight :). We will probably want to add tests to it, otherwise the chances of it breaking after we upgrade some dependency is big, and because we have so many games, automatic testing is important

flamecoals:
But I can help adding the tests for you too, I will follow up wiht more info tonight

PR #485, opened 2020-04-10
Add visual cue of whose turn it is, add test for ScoreBadges

### LINKED ISSUE #483: Add visual cue of who's turn is it
Especially in take 6 that we can have up to 10 players, it is kinda annoying to not know who should be playing.

We already have this info on ctx, just need to change the score chips to add an icon if it is user's turn...

jasonharrison:
Add visual cue of whose turn it is, add test for ScoreBadges

Closes #483


src/games/common/ScoreBadges.tsx
```diff
@@ -9,26 +9,40 @@ interface IScoreBadgesProps {
...
   colors?: string[];
+  currentPlayer?: string;
 }
 
 export class ScoreBadges extends React.Component<IScoreBadgesProps, {}> {
   render() {
     const badges = this.props.scoreboard.map((score) => {
       const nickname = this.props.players.find((player) => player.playerID.toString() === score.playerID).name;
       const isSelf = score.playerID.toString() === this.props.playerID;
+      const isCurrentPlayer = this.props.currentPlayer === score.playerID;
```
flamecoals, line 20: This wont work for takesix or other games where you can have multiple people taking actions. We should look at both ctx.activePlayers and, only if activePlayers is null, look into currentPlayer

PR #488, opened 2020-04-10
Remove email button from GameSharing due to lack of use

jasonharrison:
Remove email button from GameSharing due to lack of use


flamecoals:
FYI, if the user does not have a mail clinet installed, it also breaks their JS exxecution, so I rather not have it

PR #508, opened 2020-04-16
Add search for games

jasonharrison:
### Checklist

When entering a search query on desktop, the page shifts a few pixels, possibly because the cards have a bigger width bigger than the page width.  This issue existed prior to this PR, but it went unnoticed because the game cards on the homepage did not change dynamically (until now).


src/components/App/FBGBar/index.tsx
```diff
@@ -11,18 +11,6 @@ interface FBGBarProps {
 
 class FreeBoardGamesBar extends React.Component<FBGBarProps, {}> {
```
flamecoals, line 12: Please rename file back to FreeBoardGamesBar.ts before merging

PR #517, opened 2020-04-24
Refactoring secret code so it does not uses stages anymore

flamecoals:
Because I changed the way we do random shuffling and etc, the tests unit tests are broken. I will fix it tomorrow


flamecoals:
As discussed here: https://github.com/nicolodavis/boardgame.io/issues/638

PR #528, opened 2020-05-01
hangman rota improved

gk-patel:
Small changes, 

- hangman hint fixed 
- rota gameover board

### DIFF
--- a/src/games/rota/Field.tsx
+++ b/src/games/rota/Field.tsx
@@ -9,6 +9,7 @@ const CIRCLE_RADIUS = BOARD_SIZE * 0.4;
 const POINT_SELECTION_ZOOM = 1.2;
 
 interface IFieldProps {
+  boardSize: number;
   points: Point[];
   clickPoint: (pointID: number) => void;
   idSelectedPoint: number;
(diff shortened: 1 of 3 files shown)


src/games/rota/Field.tsx
```diff
@@ -22,18 +23,38 @@ const playerColors = {
...
     return (
-      <svg width="100%" height="100%" viewBox="-250 -250 500 500" className={css.RotaField} pointerEvents="visible">
-        <g strokeWidth="4" stroke="#cccccc">
-          {this._drawLinesAndCircle()}
-        </g>
-        <g>{this._placePieces()}</g>
-      </svg>
+      <div style={{ textAlign: 'center' }}>
+        <svg
+          width={`${this.props.boardSize}%`}
```
flamecoals, line 28: You have BOARD_SIZE and props.boardSize, what is the difference?

src/games/rota/board.tsx
```diff
@@ -80,20 +80,33 @@ export class Board extends React.Component<IBoardProps, {}> {
     }
   }
 
+  _getBoard(boardSize = 100) {
```
flamecoals, line 83: This seems to be an constant, so why make it a prop? Wouldn't it be better for it to just be a constant inside Field?

gk-patel: nope. there is one call with 50.

src/games/rota/Field.tsx
gk-patel: well, when the game gets over, it is called with 50% of the original size.

flamecoals: I see, can you rename one of those variables so they don't seem to mean the same thing?

You could rename the props to be boardSizePerc for percentage, or the other one for SVG_BOARD_SIZE

PR #554, opened 2020-05-12
WIP Proto

jasonharrison:
#### Checklist


fbg-server/src/proto/user.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+
+package user;
+
+
+service UserService {
+  rpc NewUser (NewUserReq) returns (NewUserRes) {}
+}
+
+message NewUserReq {
```
flamecoals, line 10: style guide... use Request and Response, add comments in each proto and field.

fbg-server/src/proto/user_pb.d.ts
```diff
@@ -0,0 +1,37 @@
+import * as jspb from 'google-protobuf';
+
+export class NewUserReq extends jspb.Message {
```
flamecoals, line 3: this seems auto-generated? do we need to commit it? cant it be part of our build steps?

flamecoals: if we need to commit it, just make sure we have a folder that makes it clear it is autogen

fbg-server/src/proto/UserServiceClientPb.ts
```diff
@@ -0,0 +1,55 @@
+/**
```
flamecoals, line 1: .

flamecoals:
We decided to use GQL instead, closing this

PR #557, opened 2020-05-15
Avoid needlessly passing props

jasonharrison:
#### Checklist

### DIFF
--- a/web/src/misc/UaHelper.ts
+++ b/web/src/misc/UaHelper.ts
@@ -1,9 +1,9 @@
 import MobileDetect from 'mobile-detect';
 
-export function processUserAgent(userAgent: string) {
+export function uaIsMobile(userAgent: string) {
   const md = new MobileDetect(userAgent);
   const isMobile = !!md.mobile() && !md.tablet();
-  return { userAgent, isMobile };
+  return isMobile;
 }
 
 export function isMobileFromReq(req: any) {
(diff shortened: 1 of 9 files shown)


web/src/misc/UaHelper.ts
```diff
@@ -5,3 +5,17 @@ export function processUserAgent(userAgent: string) {
   const isMobile = !!md.mobile() && !md.tablet();
   return { userAgent, isMobile };
 }
+
+export function isMobileFromReq(req: any) {
```
flamecoals, line 9: Can't we type the req ?

web/src/misc/UaHelper.ts
```diff
@@ -5,3 +5,17 @@ export function processUserAgent(userAgent: string) {
...
+  let userAgent;
+  if (req) {
+    userAgent = req.headers['user-agent'];
+  } else if (window) {
+    userAgent = window.navigator.userAgent;
+  }
+  if (!userAgent) {
+    return false;
+  }
+  const userAgentDetails = processUserAgent(userAgent);
```
flamecoals, line 19: Add unit test for this function? Would help me understand what is going on

web/src/misc/UaHelper.ts
jasonharrison: Not sure if NextJS has typings for `req`

PR #563, opened 2020-05-21
Fix backend by differentiating between internal/external bgio endpoints

jasonharrison:
#### Checklist

### DIFF
--- a/fbg-server/src/rooms/RoomUtil.ts
+++ b/fbg-server/src/rooms/RoomUtil.ts
@@ -16,9 +16,13 @@ export function roomEntityToRoom(roomEntity: RoomEntity): Room {
   };
 }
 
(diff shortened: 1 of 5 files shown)


fbg-server/src/rooms/RoomUtil.ts
```diff
@@ -16,9 +16,13 @@ export function roomEntityToRoom(roomEntity: RoomEntity): Room {
   };
 }
 
-export function getBgioServerUrl(): string {
-  const config = process.env.BGIO_SERVERS || 'http://localhost:8001';
-  const possibleServers = config.split(',');
+export function getBgioServerUrl() {
```
flamecoals, line 19: Can you define an interface here and return this interface?

PR #571, opened 2020-05-25
Add storybook

jasonharrison:
#### Checklist


flamecoals:
This would require a new webpack config which would add complexity, so we decided to use a simple documentation instead

web/src/components/App/Game/QrCodeIcon.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import QrCodeIcon from './QrCodeIcon';
+
+export default {
+  title: 'QrCodeIcon',
+  // parameters: { notes: componentNotes },
```
flamecoals, line 6: remove

web/stories/0-Welcome.stories.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
flamecoals, line 1: can you call this folder storybook?

Issue #490, opened 2020-04-10
Some nice-to-haves for Hangman.


gk-patel:
Things to add in FBG Hangman, 

- [ ] Audio 
- [ ] More than 2 players mode 
- [ ] Show Score Board at the end 
- [ ] Keyboard Input should also be supported
- [ ] make use of BGIO secrets 
- [ ] maybe set the number of turns to 10 (turn: { moveLimit: 1,}) in the game
- [ ] better Types

@flamecoals @jasonharrison : For now, will focus on other game(s), but will come-back to this later.

PR #580, opened 2020-05-30
Add score table to Hangman.

mateusazis:
The table displays the score for each player on the GameOver screen as well as between turns.

#490 

#### Checklist

Preview:
Between turns
Screen Shot 2020-05-30 at 9 54 04 AM

End of game
Screen Shot 2020-05-30 at 9 54 22 AM

### DIFF
--- a/web/src/games/hangman/board.css
+++ b/web/src/games/hangman/board.css
@@ -1,15 +1,3 @@
-.scoreTable td, th, table {
-  border: 1px solid grey;
-}
-
-.scoreTable th {
-  font-size: 35px;
-}
-
-.currentPlayerScore {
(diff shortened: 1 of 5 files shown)


flamecoals:
FYI, some little changes I've done:
- Using the common component Scoreboard instead of a table from scratch (in order to keep UI consistency). Sorry if you didnt see this component, we are working on creating better documentation.
- I am not showing it in the middle of the game since one row would always be 0 and I think it could be a bit confusing.

flamecoals:
Here is the final screenshot:
image

mateusazis:
Even better, thanks!

PR #590, opened 2020-06-03
Improve Takesix UX

eemp:
### Updates

* Use monkey themed assets.
* Use opacity to highlight the deck or player hand depending on game phase.
* Autoplay deck select for players when there is only one option.

#### Checklist

image


flamecoals:
Wow, this looks great! :tada:  I am going to check it out now, one second :)

flamecoals:
Hey @eemp, great job on this and welcome to FBG :). We could merge as is, but I have some questions:

- Can we make the bananas bigger, and maybe another color? It seems really hard to see how many bananas in each card, especially with the red background.
- Do you have the file you used for editing/creating these assets? it would be useful if we want to make tweaks in the future, like the banana size.
- Should we change the color scheme? The way it is I am afraid we might look too similar to 6 nimmt and we want to make it clear that we are a different game. I think the usage of the monkey helps a lot on this though, otherwise we wouldn't be able to merge. What are your thoughts?
Feel free to reach out to us on discord too, I haven't seen you there :)

Thanks!

eemp:
I pulled out and tweaked the cards from a pdf that I had found available freely for print and use.  I will look to see if I can avoid the images altogether and opt for something like SVG format for greater transparency, and easier modification.

flamecoals:
That would be great! We could even potentially create a SVG manually based
on what you have. SVG should give much better resolution while reducing a
lot the payload too, from >1MB to only a few hundred kbs.

eemp:
Updated the cards.

Sorry about the linting - didn't notice the target under web/package.json.

flamecoals:
This looks great! Lets just the CI finish and we can merge! Thank you so much for improving our most played game :D

flamecoals:
Btw, @jasonharrison is working on improving the development experience by running the linter on a pre-commit hook, so we wont have to wait for the CI to fail

flamecoals:
@eemp Thanks for your contribution! This was pushed to prod now.
I was playtesting with @jasonharrison and because he is color blind, he wasn't able to play with the color scheme that we merged originally. So I commited some changes to the cards and color scheme to have better contrast.

eemp:
Nice, I like the new colors better.

PR #596, opened 2020-06-09
Secret Codes: Merge giveClue with Play phases

### LINKED ISSUE #543: [Secret codes] Merge giveClue with Play phases
Playing with friends, I noticed a lot of times the spymaster would forget to click on "done", and also we are running into some weird issues where people cant click on "done". This would fix both issues.

jasonharrison:
Secret Codes: Merge giveClue with Play phases, close #543 
#### Checklist

### DIFF
--- a/web/src/deprecated-bgio-ui/2d/token.test.js
+++ b/web/src/deprecated-bgio-ui/2d/token.test.js
@@ -14,6 +14,9 @@ import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new Adapter() });
 
+// avoid error: The tag <g> is unrecognized in this browser.
+jest.spyOn(console, 'error').mockImplementation(() => {});
+
 test('animation', () => {
   const token = Enzyme.shallow(
     <Token x={1} y={2} animate={true}>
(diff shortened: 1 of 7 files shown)


web/src/games/secretcodes/PlayBoard.test.tsx
```diff
@@ -15,6 +15,7 @@ function render(client: Client, state: Any, gameOver = false) {
       moves={client.moves}
       playerID={'0'}
       isActive={true}
+      isHost={false}
```
flamecoals, line 18: What does isHost do?

web/src/deprecated-bgio-ui/2d/token.test.js
```diff
@@ -14,6 +14,9 @@ import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new Adapter() });
 
+// avoid error: The tag <g> is unrecognized in this browser.
+jest.spyOn(console, 'error').mockImplementation(() => {});
```
flamecoals, line 18: I don't think it is a good idea, it will make it hard to debug and we might miss real bugs. Can you undo?

flamecoals: Ohhh nevermind , I didn't see it was only in this test

PR #597, opened 2020-06-09
Handle disconnects

jasonharrison:
#### Checklist


flamecoals:
Create internal modules for Fbg-server:
- One "FbgJwt", where we wrap all the decode/encode JWT logic in it.
- One "FbgSubscriptions", where we wrap all the subscription shutdown and authentication in it.

On the FE, break LobbyService into AuthService and RoomService.

PR #601, opened 2020-06-15
Lobby UI Elements

jasonharrison:
#### Checklist


flamecoals:
TODO:
- [ ] Fix frontend error when getting updates from the subscription
- [ ] Create button and dialog for creating new public room
- [ ] Fix bug on desktop carousel with arrows being enabled even though the content is smaller than the area
- [ ] Redirect to the room page when the user clicks on the room button
- [ ] Add some tests

jasonharrison:
@flamecoals 

What are your thoughts on removing the "no public room" text?  On one hand, it draws attention to the fact that users can create a public room.  On the other hand, however, we have a header for "Public Rooms" and no rooms to show...

image

---

**To this (text removed):**

image

---

Secondly, the "new room" button seems to hang over slightly to the right of cards:

image

---

Lastly, I worry the naming of rooms may be somewhat confusing if you are a user that doesn't know what's going on.  (Room: Snake?  What's this mean!?)  Any thoughts on how we could communicate this and make it less confusing for users?

jasonharrison:
I think I fixed the "New Room" button hanging over.

While doing so, I noticed the UI seems a little "cramped" on a mobile display:

image

flamecoals:
@jasonharrison thanks for the fix! Regarding not having a message when there is no rooms, I dont think thats a good idea as it might seem "broken" to the user... I think we are good to merge now :)

PR #623, opened 2020-07-29
Fixed Numbering Throughout Zoo Parade

DanielSherlock:
The numbers on cards and stuff should display as 1-5 to the user, not 0-4. #620 went some way towards this, but doesn't seem to have had the desired effect. This change fixes this, the way I assumed it would be done when I was working on the code before. I have reverted the internal logic to use 0-4, made sure cards and hints only _display_ 1-5.

This means:
 - The large numbers on the cards now show 1-5 (even after #620, they showed 0-4).
 - Hints aren't kinda broken (they were before, become some of the logic was using 1-5 while the rest of it still expected 0-4).

The fact that the internal logic used 0-4 is an wierd characteristic of the code that should probably be documented to that people don't make the intuitive - but sadly buggy - changes as in #620 (thankfully @Spooky-0 warned me about this when I was working on it originally). Suggestions on how best to do this documentation are very welcome. (The alternative, of course, is a much much more intensive rewrite of the code to use 1-5 throughout, but I'm not sure it's worth it.)

Also, I have no idea about tests and stuff, so if this breaks tests (though really I don't see why it would) I'll need help unbreaking them.

#### Checklist

### DIFF
--- a/web/src/games/zooparade/components/bcard.tsx
+++ b/web/src/games/zooparade/components/bcard.tsx
@@ -26,8 +26,8 @@ interface InnerWrapper {
 
 export class BCard extends React.Component<InnerWrapper, {}> {
   render() {
-    var cardDisplayValue: string;
-    var image: any;
+    let cardDisplayValue: string;
+    let image: any;
     if (this.props.empty !== null) {
       // No Real Card Face
       cardDisplayValue = '';
--- a/web/src/games/zooparade/components/bcard.tsx
+++ b/web/src/games/zooparade/components/bcard.tsx
@@ -26,11 +26,11 @@ interface InnerWrapper {
 
 export class BCard extends React.Component<InnerWrapper, {}> {
   render() {
-    var cardValue: string;
+    var cardDisplayValue: string;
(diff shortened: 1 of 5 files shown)


web/src/games/zooparade/components/bcard.tsx
```diff
@@ -26,11 +26,11 @@ interface InnerWrapper {
 
 export class BCard extends React.Component<InnerWrapper, {}> {
   render() {
-    var cardValue: string;
+    var cardDisplayValue: string;
     var image: any;
```
flamecoals, line 30: Can you please change these "vars" to "let's"? Vars are globals and are a bad idea overall

DanielSherlock:
Thanks for the advice about using the formatting and how to run the tests locally - it turns out that was exactly the issue. Everything seems to be happy now.

I had to add the .gitattributes file else git-on-windows changes all the line-endings to CRLF, which the prettier rejects. Now, even on Windows, all the line-endings stay as LF.

flamecoals:
Thank you!

PR #624, opened 2020-07-30
Zoo parade smarter hints

DanielSherlock:
Making the hints smarter. There are two new (additional) behaviours:

1. Ruling out alternatives when recieving a hint:
That is, `[? N ? ? ?] --(correct hint 4)--> [N N N Y N]`.

2. Deducing a hint when everything else is ruled out:
That is, `[N N ? ? N] --(incorrect hint 4)--> [N N Y N N]`.

(In the above examples, `Y = IHintMask.YES`, `N = IHintMask.NO`, and `? = IHintMask.UNKNOWN`.)

The behaviours are implemented, and seem to work well. What remains is to update the tests, as they are still expecting the old behaviour. I'm unsure exactly how to do this... I'll probably leave things like this for today.

#### Checklist


DanielSherlock:
With this, both new behaviours are checked for in the tests.

web/src/games/zooparade/moves.ts
```diff
@@ -187,20 +187,17 @@ export function moveHintValue(G: IG, ctx: Ctx, IDPlayer: number, IDHintValue: nu
     hands: G.hands.map((hand: IHand, index: number) => {
       if (!(index === IDPlayer)) {
         return hand;
+      } else {
+        return <IHand>{
+          ...hand,
+          hints: hand.hints.map((hint: IHint, indexHint: number) => {
```
flamecoals, line 193: You can replace 

```Foo.map((bar) => { return baz; ...```

With

```Foo.map((bar) => (baz...```

PR #626, opened 2020-09-14
Game cardtable

cwatsonc:
#### Checklist

* [X ] Uses a separate branch feature/cardtable
* [X ] Test coverage is missing, just testing some primitive concepts/ WIP

contributing guidelines is out-of-date -- yarn run autopre and yarn run pre no longer exist in package.json


web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,48 @@
...
+        stock: [],
+
+    }),
+
+    moves: {
+        plusone(G: IG) {
+            return { count: G.count + 1 };
+        },
+
+        cutDeck(G: IG, depth: number) {
```
flamecoals, line 28: this needs to be
`cutDeck(G: IG, ctx:Ctx, depth: number) {`

web/src/games/cardtable/CardTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+    display: 'flex',
+    flexWrap: 'nowrap',
+    justifyContent: 'center',
+    alignContent: 'center',
+    backgroundColor: purple['100'],
+  },
+}));
+
+const CardTable: FunctionComponent<ICardTableProps> = (props: ICardTableProps) => {
+  const classes = useStyles();
```
flamecoals, line 58: Please move all these classes to its own "CardTable.css" file. Then, import it like so:
``import classes from './CardTable.css'``

This will remove the CSS bloat of this file

web/src/games/cardtable/CardContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+ *     title: 'Image',
+ *     author: 'author',
+ *   },
+ *   {
+ *     [etc...]
+ *   },
+ * ];
+ */
+const CardContainer: FunctionComponent<ICardContainerProps> = (props: ICardContainerProps) => {
+  const classes = useStyles();
```
flamecoals, line 66: Ditto here, move to its own CSS file.

web/src/games/cardtable/game.test.ts
```diff
@@ -0,0 +1,231 @@
+import { Client } from 'boardgame.io/client';
+import { CardTableGame } from './game';
+
+describe('deck moves', () => {
+  test('cutDeck(1) should rotate deck by one card', () => {
+    let client = Client({ game: CardTableGame });
+    client.moves.cutDeck(1);
+    const { G } = client.store.getState();
+    expect(G.deck[0]).toEqual({ id: 'AC', rank: 26, faced: false, img: './media/png/AC.png' });
```
flamecoals, line 9: The game rules (game.ts) should be independent of implementation details. The filepath is a implementation detail, so please remove the field "img" from this type and derive it when needed from the id.

web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,194 @@
...
+  south: IHand;
+  east: IHand;
+  west: IHand;
+}
+
+export interface ICard {
+  id: string;
+  rank: number;
+  faced: boolean;
+  img: string;
```
flamecoals, line 23: remove this field, it can be derived from id.

web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export interface IHand {
+  private: ICard[];
+  held: ICard[];
+  melds: ICard[];
+  tricks: ICard[];
+  played: ICard[];
+  cribFlipped?: boolean;
```
flamecoals, line 32: what is a crib?

web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,194 @@
...
+      for (var j = 0; j < pattern.players; j++) {
+        holders[j % pattern.players].push(fresh.pop());
+      }
+    }
+    G.hands.north.melds.length = 0;
+    G.hands.north.played.length = 0;
+    G.hands.north.private.length = 0;
+    G.hands.north.tricks.length = 0;
+    G.hands.north.tricks.length = 0;
+    G.hands.north.held.length = 0;
```
flamecoals, line 92: Can you overwrite the length of an array? Shouldn't it be better to do `G.hands.north.held = []` ?

web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,194 @@
...
+  private: ICard[];
+  held: ICard[];
+  melds: ICard[];
+  tricks: ICard[];
+  played: ICard[];
+  cribFlipped?: boolean;
+}
+
+export interface ILocation {
+  container: string;
```
flamecoals, line 36: instead of using a string and spliting each part with a dot, can you use an array instead?

web/src/games/cardtable/game.ts
```diff
@@ -0,0 +1,194 @@
...
+    return;
+  },
+
+  flipCrib(G: IG) {
+    G.hands.east.cribFlipped = !G.hands.east.cribFlipped;
+  },
+};
+
+const getNamedContainer = (G: IG, name: String) => {
+  let path: string[] = name.split('.');
```
flamecoals, line 138: Use an array instead of splitting a string. It is a bad practice to use a string to store structured data.

flamecoals:
Thank you for your contribution! Having this game will be a great addition to FBG and I am sure users will love it :).
Some questions:
- Are the rules coded of Cribbage? Because this is a classic game, you could name your game "Cribbage", and it will help bring users from search engines. If you want to create other games in the future using the same UI and etc, you can re-use some components and still have the rules coded for each game. 
- How are you planning for users to trigger actions ? I couldnt find buttons for triggering the actions available on the debug menu.
- Can you move the card component to gamesShared? I think this would be a useful component for other games too!
- Can you create some stories for your components? This is a form of documentation that will help other developers to understand and help with your game. As an example, you can check any file finishing with "storybook" on the name on other games folders. 
- Are you using the blue background for debugging? It would be better if it was black in the final version.
- The horizontal and vertical scrolling inside each container is probably not going to be something users are going to find easy to use. It would be better if we only allowed horizontal scrolling for each element, and a single vertical scroll for the whole page. Check this that might help:
https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--mobile
https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--desktop


image

web/src/games/cardtable/game.test.ts
cwatsonc: This is a test scenario, the file is a utility for loading game state in the game.test.  

The data model for the game is still in flux as many parts will be pared away -- would like to defer the decision on the card.ID property as the name is coincidental and not purposely matching the glyphs for the deck (which is a free distribution from American Contract Bridge League (ACBL) don't know if they're appropriate for copyleft distribution.

web/src/games/cardtable/game.ts
cwatsonc: This one needs to remain as implemented for now -- the game currently has no abstraction for private cards, in my earlier versions of this game the cards were given anonymous identities to help prevent cheating on the client.  The server was the authoritative glue between anonymousID -> cardGlyph and managed privacy filters for each players context.  card.ID will likely become this mechanism. and at that point I don't want to rip a bunch of associations out that calculate that mapping.  If this isn't a strong objecting I ask for some grace on this early development.

web/src/games/cardtable/game.ts
cwatsonc: The act of 'laying-to-crib' is a reference to that relationship and laying your cards into a place where they're kept in secret during cribbage play; the dealer and the opponent(s) both lay to "the crib" secretly and the dealer is given an opportunity to score on their value as the final phase of hand play.

Cribbage takes it name from the practice of laying up ships/boats in cribbing when in dry dock.  A crib is also a childs bed shaped like those dockside affairs \__/

web/src/games/cardtable/game.ts
cwatsonc: The persistence docs for Immer state that this is one of the only properties you can change on an array in their framework.  I believe this fragment is performing a reset during move.deal( ) it's pretty direct and easy to understand, but if we've got a style we're after I'm game.  One topic came up that the card abstractions should be moved to common -- I don't disagree.  While in dev let's give it some time to breathe before we start tearing apart.  Many of my solutions are first time efforts -- this is a big lift for my tech skills and I'm trying to be patient while I learn the ropes.

web/src/games/cardtable/game.ts
cwatsonc: again this is a first attempt before understanding the total of constraints the design will require.  I'm not opposed to naming conventions that are checked in automation (linting) and well expressed in the docs.  I needed a way to traverse the object graph that was readable and easy to understand.  I'm not sure splitting the path into multiple keys achieves that, but willing to look at suggestions. Again this is a primitive card abstraction and is likely to move to common (or be replaced by a better version).

web/src/games/cardtable/game.ts
cwatsonc: So, since I'm still learning my chops with JS you're going to see a lot of mealy code--be patient and I'll be open to moving my solution so it looks like what you're used to.  I'm not sure that:

keysPath = ['mycontainer', 'below','cards','blender'] is more understandable than 'mycontainer.below.cards.blender' but I'll move this into my next release as an issue.  I was asking you folks for guidance as much as defending my code line.  I'm happy with the progress and am getting comfortable with the data model -- this app may not fully represent your standards, but it would make a nice tutorial -- the state model and ui are very simple which makes a good starter and the abstractions are very familiar to most of us -- table, deck, hand, deal, cut, all of them built around just a couple of very sturdy primitive mutators -- if this core is moved to common my game is about 200 lines of JS.  I'm going to eventually have more Jest/Enzyme coverage as the UI matures, but the game has very few lines that aren't covered.

web/src/games/cardtable/CardContainer.tsx
cwatsonc: > * Are the rules coded of Cribbage? Because this is a classic game, you could name your game "Cribbage", and it will help bring users from search engines. If you want to create other games in the future using the same UI and etc, you can re-use some components and still have the rules coded for each game.

The Card Table Game can be named **anything**, but it isn't a Cribbage game it's the first version of a card game simulator.  It's designed to support remote gameplay for the cumbersome abstract operations you execute when playing at home.  Shuffle deal, play tricks, quit tricks, are all operations that are exposed to the client.  To the extent that it won't impede players who wish to adapt their own style of play it will have many options to strip away the computer guardrails on rules.  Cribbage is a family favorite but we regularly meet and play three or four card games during a visit.  Think of cardtable as an attempt to facilitate such outings in an online community.

> * How are you planning for users to trigger actions ? I couldnt find buttons for triggering the actions available on the debug menu.  

There are more moves in the tip of feature/cardtable waiting to be merged back to you.  Currently it supports moveCard, cutDeck, deal, putToCrib, play, pegPoints, flipCrib. I played a full game in a test this evening and was pleased at how fluid the play was -- state diagrams mapping out the game are my next step -- there's a couple easy ones and many complicated but as I've mentioned these are early steps in a longer journey.  


> * Can you move the card component to gamesShared? I think this would be a useful component for other games too!
I will as the community agrees to the solution and finds value in it.  It's a labor of love, not my main line of development.  I mentioned one of your big successes is you remain approachable -- the barriers to integrating into such an effort are steep if you pick the wrong community.

> * Can you create some stories for your components? This is a form of documentation that will help other developers to understand and help with your game. As an example, you can check any file finishing with "storybook" on the name on other games folders.

I took a dive into Storybook and will try and keep my docs up-to-date.  I added a more representative thumbnail and a good deal about cribbage play on my description.md.  If you see something that needs improving on those first attempts give me some specific feedback.  I've scrolled through most of the Storybook collection, I can't see any details about any of the previous submissions.  Point out a good example and I'll try and emulate it -- All of the ones I encountered on the document set were pretty plain descriptions of components -- we also appear to be using only 1/2 of the framework -- there's a canvas tab on each of the doc nodes that is empty.  The doc framework looks interesting -- point me in that direction with some good docs, and I'll provide the lightest that we can agree to.


> * Are you using the blue background for debugging? It would be better if it was black in the final version. 

I am using color to help with my struggles in doing html/js/react.js ui solutions -- I have a hard time with CSS and it sometimes takes simplifications to see subtle problems with layout and display characteristics -- I learned the color tricks from a course on Web Dev by Colt Steele. I find it a good tool to debug UI.  Card tables are green usually, and that is one consideration, but a late one -- UI is decoration for the app.
  
> * The horizontal and vertical scrolling inside each container is probably not going to be something users are going to find easy to use. It would be better if we only allowed horizontal scrolling for each element, and a single vertical scroll for the whole page. Check this that might help:
>   https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--mobile
>   https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--desktop

Thanks, I'll look at those... I found the carousel docs in the Storybook offerings.  I couldn't tell if it was adaptive and responsive by design -- React.js has been in my sights for a while and while I'm still learning you'll see my experiments with various web design techniques.  Most of my React.js so far has been functional components, the same guy(s) who taught the web bootcamp (Colt Steel) is teaching me about a bunch more, contexts, hooks, redux, router. It's just a lot to get in one gulp -- but I like the component architecture a lot I was experimenting with components when a colleague asked me to investigate React.js -- I find the opinionated frameworks easier to learn, as long as you don't stray too much from their bread and butter, React uses semantic-ui natively (I think) and you guys are using another CSS framework.  That kind of impedance snowballs over time, it would be good to be declarative, but also provide the option of customization -- "batteries included, but optional" 

It's like drinking from Niagra Falls most days.  I planned on correcting the UI behavior after the state graph model was delivered.  Still a lot outstanding aside from the UI/UX.

Thanks for a great review -- recommend we move standard concerns to linter logic -- docs-- where new folks will appreciate the guidance and guardrails.  If I forget to get you the requested improvements please open issues so I can get the broader opinion of the community as I learn.

web/src/games/cardtable/game.test.ts
flamecoals: but in general you dont want a file path as part of your game state. The game state is stored on databases after a match is played. What if you need to change the file path later, or change the image? You would have the stale file path stored on the database for all past games, and it would break the game for users coming back to them.

If ID is not enough, create another enum or something that would be enough. But **do not** store filepaths here.

web/src/games/cardtable/game.ts
flamecoals: See discussion above. File paths are not OK here.

web/src/games/cardtable/game.ts
flamecoals: This isn't about naming conventions, it is about using the language correctly. Strings are meant to store text, not data structures.

I won't block your game on this, but using dividers and splitting text to store data structures is a really bad practice.

web/src/games/cardtable/game.ts
flamecoals: using keysPath = ['mycontainer', 'below','cards','blender']  isnt storing a deep object, just a reference in a data structure that actually means "several strings" instead of one . I wont block you on this, but it is a bad practice.

web/static/sitemap.xml
```diff
@@ -0,0 +1 @@
...url><loc>https://www.freeboardgames.org/about</loc></url><url><loc>https://www.freeboardgames.org/play/estatebuyer</loc></url><url><loc>https://www.freeboardgames.org/play/mancala</loc></url><url><loc>https://www.freeboardgames.org/play/secretcodes</loc></url><url><loc>https://www.freeboardgames.org/play/takesix</loc></url><url><loc>https://www.freeboardgames.org/play/checkers</loc></url><url><loc>https://www.freeboardgames.org/play/fourinarow</loc></url><url><loc>https://www.freeboardgames.org/play/seabattle</loc></url><url><loc>https://www.freeboardgames.org/play/tictactoe</loc></url><url><loc>https://www.freeboardgames.org/play/hangman</loc></url><url><loc>https://www.freeboardgames.org/play/zooparade</loc></url><url><loc>https://www.freeboardgames.org/play/chess</loc></url><url><loc>https://www.freeboardgames.org/play/cornerus</loc></url><url><loc>https://www.freeboardgames.org/play/reversi</loc></url><url><loc>https://www.freeboardgames.org/play/tictactoeplus</loc></url><url><loc>https://www.freeboardgames.org/play/rota</loc></url><url><loc>https://www.freeboardgames.org/play/ninemensmorris</loc></url><url><loc>https://www.freeboardgames.org/play/cardtable</loc></url></urlset>
```
flamecoals, line 1: Please delete this file.

cwatsonc: @flamecoals done in latest commit #626

web/src/games/cardtable/game.ts
cwatsonc: No updates as of yet.  Still tracking this request 

> Use an array instead of splitting a string. It is a bad practice to use a string to store structured data.

Can you explain what you mean by use an array rather than splitting?  I'm going to take a long look at the game after it deploys latest changes.  I'm willing to implement this function any which way we deem.  The other factor I'm considering is the request to pull standard components to common -- I think that's a great idea and it would save a lot of work -- work on this product in advance of that moment should be minimal.  If it's good enough for WIP deployment i would appreciate the lift from the community at refining the approaches as we see fit.

flamecoals:
Ok, let's merge this because you made good progress and others can help. But takeaways from the review:

- Avoid filepaths on the game state (G), as it is stored on the database and will have versioning/stale data issue and cause bugs. This **must** be done before publishing the game.
- Try to avoid relying on strings too much, use ['mycontainer', 'below','cards','blender']  instead of 'mycontainer.below.cards.blender'...

flamecoals (approved):

PR #642, opened 2020-10-24
Showing debug panel in production

flamecoals:


flamecoals:
Reference: https://boardgame.io/documentation/#/debugging?id=using-the-debug-panel-in-production

cwatsonc (approved):
Change looks fine and is approved.  Need notes on where the flag need be set by game developer if they want this enabled.

PR #655, opened 2020-11-05
Chat

flamecoals:


web/src/infra/chat/internal/ChatMessageHistory.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import ChatMessage from './ChatMessage';
+import Message from './Message';
+
+interface Props {
+  messages: Message[];
+}
+
+export default function (props: Props) {
+  return <>{props.messages.map(_createMessage)}</>;
```
jasonharrison, line 10: I would do this a little differently:

```typescript
export default function ChatMessageHistory(props: Props) {
  return props.messages.map(message, idx) => (
    <ChatMessage key={idx} message={message} />
  ));
}
```

Two things:
1. I added a name to the functional component.  Anonymous function declarations can cause various issues, like causing React to needlessly re-render.  When React receives a new anonymous function (e.g. on hot reload), it cannot be compared to the previous anonymous function.
2. I added a `key` prop since we're iterating to help React handle the minimal DOM change.

jasonharrison (approved):
Nice!  I left you one small comment to check out, but good work!

web/src/infra/chat/internal/ChatMessageHistory.tsx
flamecoals: Done!

PR #665, opened 2020-11-09
Add main game tests for Mergers

lfkellogg:
Includes unit tests for the moves, and then three scenario tests including one that plays out an entire game on a small board.

#### Checklist

### DIFF
--- a/web/src/games/mergers/game.test.ts
+++ b/web/src/games/mergers/game.test.ts
@@ -1,0 +1,1220 @@
+import { Ctx } from 'boardgame.io';
+import { INVALID_MOVE } from 'boardgame.io/core';
+import {
+  autosetChainToMerge,
+  awardBonuses,
+  buyStock,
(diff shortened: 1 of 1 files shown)


flamecoals:
This looks good. Just a tip: you can instantiate a Boardgame.io client for unit tests too, this way you can avoid initializing the default state by hand and also test more complex behavior closer to what the user will do (i.e. that given actions by several players will get the Ctx to some state). Documentation here:
https://boardgame.io/documentation/#/testing?id=testing-strategies

A good example is here: https://github.com/freeboardgames/FreeBoardGames.org/blob/0c5d8ef0ebbf68b5a804618384aba6e5976f9e1f/web/src/games/seabattle/game.test.tsx#L11

But this looks good, just something to keep in mind for the future :)

flamecoals (approved):

lfkellogg:
@flamecoals Thanks for the tip!

PR #666, opened 2020-11-09
Updates helm replicas

flamecoals:
#### Checklist


cwatsonc (approved):
these look fine to me.  Is there a method to define a default if these yaml props are undefined?

flamecoals:
> these look fine to me. Is there a method to define a default if these yaml props are undefined?

Values.yaml are the default. Users can provide using the -f flag a new yaml that may overwrite only a few of these attributes.

PR #673, opened 2020-11-15
Dracula6

Spooky-0:
#### Checklist

* [ ] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).


web/src/games/secretDracula/board.tsx
```diff
@@ -553,4 +556,13 @@ export class Board extends React.Component<IBoardProps> {
       return;
     };
   }
+
+  _getGameOver() {
+    // Online game
+    if (this.props.ctx.gameover.win == true) {
```
flamecoals, line 559: By looking at this implementation, I suspected this code wouldn't work, and it seems like thats the case:
image
I will fix it for you though, should be a simple fix.

flamecoals (approved):

PR #674, opened 2020-11-17
Add Mergers UI

lfkellogg:
This is the full working UI for Mergers. It's in pretty good shape with some room for improvement in a few areas.

Here's how I think it's doing against the release checklist:

- [x] Does not break intellectual properties of others. Check for copyright in game name, instructions, graphics, etc.
  - Only thing from the original game is the name of the hotel chains ("Tower", "American" etc.) which I'm happy to change if needed.
- [x] Playable end-to-end on desktop devices.
- [x] Playable end-to-end on mobile devices.
- [ ] Has big enough touch targets (buttons, etc) on mobile.
  - This is borderline, curious to hear what other folks think. Tough to improve it much without introducing pan/zoom, although we could try rotating the grid 90 degrees on mobile.
- [x] Has high test coverage for the rules, and ideally some tests for the UI.
  - Good coverage of the rules. Decent coverage of the UI but I'd eventually like to add some more.
- [ ] Shows the game over page correctly.
  - I have a custom game over page. I see there's a shared component - should I try and use that one?
- [x] Has appropriate thumbnail, instructions, etc.

#### Screenshots

Mobile: 
image
image
image

Desktop: 
image
image
image

#### Checklist

### DIFF
--- a/web/src/games/mergers/components/HotelGrid.tsx
+++ b/web/src/games/mergers/components/HotelGrid.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import StarIcon from '@material-ui/icons/Star';
+import ClearIcon from '@material-ui/icons/Clear';
+
+import { Chain, Hotel } from '../types';
+import css from './HotelGrid.css';
+import { Hotels } from '../hotels';
+
+interface HotelGridProps {
+  hotels: Hotels;
+  lastPlacedHotel?: string;
+  isPlacingHotel: boolean;
+  playerID: string;
+  onHotelClicked: (id: string) => void;
+}
+
+interface HotelGridState {
+  hoveredHotel?: string;
+}
+
+export class HotelGrid extends React.Component<HotelGridProps, HotelGridState> {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  hotelClass(hotel: Hotel) {
+    if (hotel.hasBeenPlaced) {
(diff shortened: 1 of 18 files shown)


flamecoals:
This code is looking great, very readable and well structured. Great job!

I am going to run locally to check the tap target issue. But we could sprinkle some CSS magic to increase them in mobile devices only. We could also leverage the Carousel component to fit more things on the screen:
https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--mobile

web/src/games/mergers/components/HotelGrid.tsx
```diff
@@ -0,0 +1,108 @@
...
+          {i + 1}
+        </td>,
+      );
+    }
+    return <tr>{headers}</tr>;
+  }
+
+  render() {
+    const chainTiles = {};
+    for (const key of Object.keys(Chain)) {
```
flamecoals, line 88: I didn't know this worked for enums, TIL :) !

flamecoals:
You also did class correctly, not storing a un-serializable class on the G, and only storing an interface instead.  You avoided a common pitfall I see all the time.

flamecoals (approved):

lfkellogg:
Thanks for the review! I'll check out that carousel component.

> You also did class correctly, not storing a un-serializable class on the G, and only storing an interface instead. You avoided a common pitfall I see all the time.

As you can imagine I did try just storing the class directly in the G at first. I had to learn that lesson the hard way :)

Issue #272, opened 2019-06-09
[Infrastructure] Spectate an ongoing game


jasonharrison:

sahajanand-games:
I can try my hand at this.

gk-patel:
I can give it a try

gk-patel:
![image](https://user-images.githubusercontent.com/42677721/101920990-c4706700-3bcc-11eb-8d9b-f1d900076ed9.png)
maybe this can be handled better. If the players press the play-again button, technically, the spectators could either be redirected to the main page OR to the new session. 
Whatever is easy .... but the error message should go away.

gk-patel:
![image](https://user-images.githubusercontent.com/42677721/101927176-75c6cb00-3bd4-11eb-8e1a-5e3e4ba5514a.png)
This is what you get when you try to go into full ROOM and the game has not started.

flamecoals:
Hey @gk-patel, do you mind opening two other issues for these things you pointed out? Im going to close this for now because we already support spectating ongoing games

gk-patel:
done!

PR #676, opened 2020-11-19
Enabling spectator view for tic-tac-toe

flamecoals:
#### Checklist

### DIFF
--- a/web/src/games/tictactoe/board.tsx
+++ b/web/src/games/tictactoe/board.tsx
@@ -11,8 +11,7 @@ import { IGameArgs } from 'gamesShared/definitions/game';
 import { GameLayout } from 'gamesShared/components/fbg/GameLayout';
 import { Circle, Cross, Lines } from './Shapes';
 import Typography from '@material-ui/core/Typography';
-import { isOnlineGame, isAIGame } from '../../gamesShared/helpers/gameMode';
-import { isSpectator } from 'gamesShared/helpers/GameUtil';
+import { isFirstPersonView } from 'gamesShared/helpers/GameUtil';
 
 interface IBoardProps {
   G: any;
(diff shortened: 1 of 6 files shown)


flamecoals:
Related: #272 

How to spectate:
- Create a room, select tic-tac-toe
- Start the match
- Share the link with spectators

We would need to go game by game and add logic for rendering each game correctly for spectators. This is done here for tic-tac-toe as an example. Maybe @gk-patel can help :)

web/src/games/tictactoe/board.tsx
```diff
@@ -53,7 +53,7 @@ export class Board extends React.Component<IBoardProps, {}> {
   }
 
   _getGameOver() {
-    if (isOnlineGame(this.props.gameArgs) || isAIGame(this.props.gameArgs)) {
+    if ((isOnlineGame(this.props.gameArgs) || isAIGame(this.props.gameArgs)) && !isSpectator(this.props.playerID)) {
```
jasonharrison, line 56: Can this expression be turned into a helper method?  It's long and also duplicated here https://github.com/freeboardgames/FreeBoardGames.org/pull/676/files#diff-d09ac02e89c56265e00543c24377fb3a2849f155f794c688c40887111bd1e0caR39

jasonharrison (approved):

web/src/games/tictactoe/board.tsx
flamecoals: Yeah, I was thinking the same thing. But I don't know how to describe it. I think it should be "isFirstPersonView", maybe ? Because the other methods are watching the game as a "third-person". Do you have any ideas ?

flamecoals: It is also duplicated in other games that support all modes (chess, for instance)

flamecoals:
Found a bug, it isnt working if the user is logged off. Fixing it now.

PR #680, opened 2020-11-20
Optimize docker image size by using alpine as base

larry801:
#### Checklist

Current docker image contains yarn package cache and unused dev dependencies.
They are redundant for running in prod.

### DIFF
--- a/fbg-server/Dockerfile
+++ b/fbg-server/Dockerfile
@@ -1,14 +1,14 @@
-# web
+# fbg-server
 FROM fbg-common:latest AS common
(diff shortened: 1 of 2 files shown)


flamecoals:
Interesting. Let me test this in staging and I merge later.

Btw, why are you cleaning the cache after install?

larry801:
It is the same as apk add --no-cache.
We can remove something useful for local environment but redundant for a docker image.
Please refer to this link for detail.
https://github.com/yarnpkg/yarn/issues/986

flamecoals (approved):

PR #681, opened 2020-11-25
Fix scrolling issue for android devices on soup of letters and seabattle

flamecoals:
#### Checklist

### DIFF
--- a/web/src/games/seabattle/board.tsx
+++ b/web/src/games/seabattle/board.tsx
@@ -82,7 +82,7 @@ export class Board extends React.Component<IBoardProps, IBoardState> {
       <GameLayout
         optionsMenuItems={this._getOptionsMenuItems}
         gameArgs={this.props.gameArgs}
-        avoidOverscrollReload={true}
+        avoidOverscrollReload
       >
         {child}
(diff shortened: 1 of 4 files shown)


web/src/games/seabattle/board.tsx
```diff
@@ -79,7 +79,11 @@ export class Board extends React.Component<IBoardProps, IBoardState> {
       );
     }
     return (
-      <GameLayout optionsMenuItems={this._getOptionsMenuItems} gameArgs={this.props.gameArgs}>
+      <GameLayout
+        optionsMenuItems={this._getOptionsMenuItems}
+        gameArgs={this.props.gameArgs}
+        avoidOverscrollReload={true}
```
jasonharrison, line 85: Just a tip I learned: if you pass no value for a prop, it defaults to `true`.  So this can be just `avoidOverscrollReload`

jasonharrison (approved):

web/src/games/seabattle/board.tsx
flamecoals: Good tip, thanks!

PR #684, opened 2020-11-27
Bombs and Bunnies - ready for publishing

b-hub:
#### Checklist

* [ ] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).


flamecoals (approved):

flamecoals:
Few questions before we publish...

- Are we missing the user nickname on top of the card ? I see a little clock emoji and I was expecting to have the nickname of the player (or Player 1 vs Player 2 on local).
- Is the white background intended? I feel like the previous yellow background was much easier on the eyes (which was the same used in the state buyer game). Or why not leave it black, as most games do ?
- On mobile, it seems like the white background is hiding the FBG bar for most phones (with the exception of very tall phones like iPhone X):
image
- This isn't a problem if we keep the black background as it would not hide the bar... Do you mind if we keep it black and just change the text to white, like below ?
image

If you agree, I can do these changes :)

flamecoals (changes requested):
LGTM but small issues on mobile, waiting for it to be resolved

flamecoals:
Also, the card on the menu with white background is weird, as it mixes with the normal background of the page, and loses the card shape:

image

flamecoals:
This is really cool btw:
image

great job :100:

b-hub:
image
I've added a wooden texture which goes some way to darkening the screen. I'm not really a fan of the black background, so I hope this works for you.
Other changes I've made:

- Add player name in status bar
- Reduced the height of the game and card stack sizes to make everything a bit more compact
- Bet/Skip buttons changed to have a solid background so they are easier to see on the new texture
- Yellow shadow box added to buttons to follow with the 'selectable' styling convention
- Thumbnail has black background

flamecoals (approved):

PR #698, opened 2020-12-04
Optimize docker image size

larry801:
#### Checklist


web/Dockerfile
```diff
@@ -1,14 +1,15 @@
 # web
 FROM fbg-common:latest AS common
-FROM node:14.5.0-buster
-RUN groupadd -g 999 appuser && useradd -m -d /appdata -r -u 999 -g appuser appuser
+FROM node:14.15.0-alpine3.12 AS builder
+RUN apk add --no-cache autoconf automake file g++ libtool make nasm libpng-dev libc6-compat
+RUN addgroup -g 938 appuser && adduser -h /appdata -S -u 997 -G appuser appuser
 RUN rm /bin/su
```
cwatsonc, line 6: just curious if this is better stated ```RUN rm -rf /bin/su```  ? I'm not sure if disk leaks are possible but this looks like it would create one.

cwatsonc (approved):
LGTM, but I'm just looking at the recent changes.  Docker file question about removing /bin/su is just my curiosity. I've never seen docker files that perform directives above the FROM directive, what does this do in practical terms?

web/Dockerfile
cwatsonc: @larry801 ignore comment about placing directives above FROM in docker file I was misreading this from the commit history shown on review.  That is also probably true of my question about dependency being moved in package.json

flamecoals:
LGTM, but before merging the PR i will do a quick sanity check in our staging environment tonight to make sure everything works

web/Dockerfile
cwatsonc: @larry801 similarly, my comments about rm /bin/su were a misunderstanding/perception that this was a directory like /usr/sbin brain is on some sort of time delay today... this looks fine.

flamecoals:
I just tested in staging, worked flawlessly! Thanks!

PR #700, opened 2020-12-07
Gkp bingo

gk-patel:
#### Checklist

# General Notes 
This is just for first review, so please keep the following in mind, 

1. Local mode is temperory, so is the debug:true flag
2. The Time out interval can be adjusted, I would suggest 8 seconds for the final version, (now is 5) 
3. How many maximum players do we want at the end, currently 20, I suggest 30 ? 
3. Every player is allowed to fires a timer, based on a backoff mechanisum, which currently has maximum backoff time of 2.5 seconds. 
4. Tests still need to be written.

## How does the backoff work ? 
In G, there is a list of activePlayers, the maximum number of players in this list is controlled by MAX_BACKOFF_CANDIDATES (see ./constant.ts). Lets, say activePlayers = [ '2', '1', '3', '7', '0'], in this case 2 goes first, 1 goes after 500 ms, 3 goes after 1s, and so on and thereafter everyone goes after 2.5 seconds. 

Why did you make it so complicated ? 
Well I did not want to have a single point of failure, because if that person leaves the game or gets disconnected, then the game is basically over. 
If I made everyone fire the timer at the same time, then all clients get into a racing condition which the BGIO server does detect but that sometimes leads to a page refresh and I did not want that.


web/src/games/bingo/board.tsx
```diff
@@ -0,0 +1,187 @@
...
+          </tspan>
+          <tspan x="50%" dy={msgLineHeight * 2}>
+            Better luck next time !
+          </tspan>
+        </text>
+      );
+    }
+
+    // calculate backOff based on the list of activePlayers
+    // first person is 0 ms, then 500 ms, ..., and the rest will have max 500*length ms
```
flamecoals, line 135: What happens if the first person internet delays 1s while the second person internet delays only 0.1s ? will this break the game?

web/src/games/bingo/board.tsx
```diff
@@ -0,0 +1,187 @@
...
+          style={{
+            ...btnStyles,
+            justifyContent: 'left',
+          }}
+          onClick={() => this.setState({ showCallTable: !this.state.showCallTable })}
+        >
+          Table
+        </Button>
+        <Button
+          key="bi_shout_btn"
```
flamecoals, line 85: This function is very large. Can you move these buttons to their own functions, private to the class?

web/src/games/bingo/components/playCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <g
+      key={`bi_card_group_${number.id}`}
+      data-testid={`bi-card-group-${number.id}`}
+      onClick={() => {
+        if (!isWildcard) {
+          props.onNumberClicked(number);
+        }
+      }}
+    >
+      <rect
```
flamecoals, line 34: This is a very large function and hard to read. Can you break down sub-components (i.e. the rect) into its own private method?

web/src/games/bingo/game.ts
```diff
@@ -0,0 +1,115 @@
...
+        found[1] = found[1] + (xPos === 1 && marked);
+        found[2] = found[2] + (xPos === 2 && marked);
+        found[3] = found[3] + (xPos === 3 && marked);
+        found[4] = found[4] + (xPos === 4 && marked);
+        // rows
+        found[5] = found[5] + (yPos === 0 && marked);
+        found[6] = found[6] + (yPos === 1 && marked);
+        found[7] = found[7] + (yPos === 2 && marked);
+        found[8] = found[8] + (yPos === 3 && marked);
+        found[9] = found[9] + (yPos === 4 && marked);
```
flamecoals, line 79: can you make this into a for loop, on its own private function?

web/src/games/bingo/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+import { MAX_BACKOFF_CANDIDATES } from './constants';
+
+export function shuffleArray(array) {
+  const _array = [...array];
+  // for (let i = _array.length - 1; i > 0; i--) {
+  //   let j = Math.floor(Math.random() * (i + 1));
+  //   let temp = _array[i];
+  //   _array[i] = _array[j];
+  //   _array[j] = temp;
+  // }
```
flamecoals, line 11: Remove commented out code

flamecoals (changes requested):

web/src/games/bingo/board.tsx
gk-patel: Well then the second person will fire before the first one and the game will continue. And thereafter the, second person will become the first person.

web/src/games/bingo/board.tsx
gk-patel: Well they needed a common div, otherewise I had it seperate. But IMHO, they are just 2 buttons and also have common styles, so they can stay.

web/src/games/bingo/components/playCard.tsx
gk-patel: No No. The rect and the text must be in sync. I think this can stay. They use common xPos and yPos, etc.

web/src/games/bingo/game.ts
gk-patel: you are right, this could have been made a loop.

web/src/games/bingo/utils.ts
gk-patel: Opps. this is some DEV thing. This needs to be UNCOMMENTED, otherwise the numbers wont shuffel.

flamecoals (approved):

web/src/games/bingo/board.tsx
cwatsonc: I've just started reviewing this code, but it seems we're either missing a facility (async scheduling) or perhaps a better way to construct it out of existing infrastructure.  With the existing design I think that you should simplify the behavior so you're only ever in danger of sending one before a failover occurs:

in state: (current assignment: player1)
in state/config (assignment failover list (player1, 2, etc...)
in state (events with time displayed)

when the even schedule shows an expired event execute the update logic by current Assignment... else find failover assignment.

gk-patel: I think my logic is a bit more evolved, it chooses a few candidates and puts them in a sequence and if all fail then it allows everyone to fire. I think what you are suggesting is that we have one candidate and if he fails, everyone tries. 
I made a lot of effort to implement this backoff mechanism, and would be a shame to downgrade.

cwatsonc: Another question I have is could a Timer be included in move logic (SetTimeout)?  This feels like a pub/sub problem. I like the authoritative state to be publisher as it keeps the chatter down on the network.

cwatsonc: > I think my logic is a bit more evolved, it chooses a few candidates and puts them in a sequence and if all fail then it allows everyone to fire. I think what you are suggesting is that we have one candidate and if he fails, everyone tries.
> I made a lot of effort to implement this backoff mechanism, and would be a shame to downgrade.

I'm not suggestion a downgrade, just a change in the failover selection being round-robin based upon a selection algorithm like you describe... I hear in your description a "shotgun mechanism" everybody tries at once. I'm still looking at the code where does this reside?

gk-patel: Please try to follow variables BACKOFF_INTERVAL and G.activePlayers. As soon as a player fires timer, he comes to the top of the list of activePlayers and if he fails the next one to fire first comes to the top, simple ? 
And the number of players who can stay in the list is controlled by MAX_BACKOFF_CANDIDATES, currently 5 which makes the probabily nearly zero of every one firing at the same time.

gk-patel: I just did a really crazy experiment, i had 6 windows open and tested, 
case 1: if all players leave, play resumes at the last player 
case 2: game continues even if only one player remains

gk-patel:
@flamecoals @cwatsonc  Bingo! is done! Fell free to merge, even tests are finished. Everything finalized.

flamecoals:
Hi @gk-patel,

the game is looking really nice! My only concern is with the thumbnail, it seems like it is an image heavily edited... Did you edit it yourself? Or did you get it from somewhere? If you did, what is the license of where you got it from, is it compatible with ours ? Can you post here the link ?

flamecoals:
I will merge for now but leave it as under development, and after your answer these concerns about copyright we can merge.

Also, the subtitle "The game needs no introduction" isn't really good for discoverability by search engines. It would be best if it had words related to bingo, like "Test your luck with Bingo" or even "Play Your Luck With Bingo".

flamecoals:
@gk-patel we cant merge because you have one test failing:
```
 FAIL  src/games/bingo/board.test.tsx (8.777 s)
  ● Bingo UI › table shows played numbers

    expect(received).toContain(expected) // indexOf

    Expected value:  13
    Received string: " Bingo 1010344714626TableBingo! ⭐⭐⭐"

      63 |     wrapper.setState({ showCallTable: true });
      64 |     for (let i = 1; i < callRef; i++) {
    > 65 |       expect(wrapper.text()).toContain(state0.G.players['0'].numbers[i].value);
         |                              ^
      66 |     }
      67 |     // the current value of call should not be shown
      68 |     expect(wrapper.text()).not.toContain(state0.G.players['0'].numbers[callRef].value);

      at Object.<anonymous> (games/bingo/board.test.tsx:65:30)
```

Sorry, Travis (CI) is having an issue right now, I contacted their support to fix it

flamecoals:
Chatted on discord: Image comes from https://www.freepik.com/free-vector/bingo-background-with-ballots-colored-balls_1188623.htm#page=1&query=bingo&position=4 ... There is attribution on the instructions, so we should be fine to merge. Just need to fix the test. And optionally change the subtitle.

flamecoals:
CI passed. merging now

PR #731, opened 2020-12-24
Feature/cardtable 

cwatsonc:
#### Checklist

* [X ] Use a separate branch in your local repo (not `master`).
* [X ] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).

Recent UX and state model improvements.  Much improved game flow.

TODO:

clean up the resource loading to remove strings from state model.
provide guard logic on UX to prevent dangerous clicks on opponents cards
provide a secret state filter per player on the server logic
video for gameplay to be updated
refine UX incorporate improvements


package.json
```diff
@@ -29,6 +29,7 @@
   },
   "devDependencies": {
     "chalk": "^4.1.0",
-    "shelljs": "^0.8.4"
+    "shelljs": "^0.8.4",
+    "cypress": "^5.4.0"
```
flamecoals, line 33: Any reason to add cypress as a dependency ? If this was accidental, it would avoid changes on the lock too, so you could revert this file and yarn.lock.

web/src/games/cardtable/CribbageBoardB.tsx
```diff
@@ -0,0 +1,222 @@
...
+  className: string;
+  score: IScore;
+  opponent: boolean;
+}
+
+function Peghole(props: PegholeProps) {
+  let clazz = css.Peghole;
+
+  const isPegged = (idx: number) => {
+    let pegged = false;
```
flamecoals, line 22: You can replace isPegged implementation with:

``return  (idx < 1 && props.score.back === idx)  ||
               (idx < 1 && props.score.front === idx) ||
               (idx < 61 && props.score.front === idx) ||
               (idx < 61 && props.score.back === idx) ||
               (props.score.front > 60 && props.score.front % 60 === idx) ||
               (props.score.back > 60 && props.score.back % 60 === idx);

flamecoals (approved):

flamecoals:
Looking great! I just left some comments,  and CI Failed with this error:
```
Failed to compile.

./src/games/cardtable/CribbageBoardB.css
Global CSS cannot be imported from files other than your Custom <App>. Please move all global CSS imports to src/pages/_app.tsx. Or convert the import to Component-Level CSS (CSS Modules).
```

All the CSS files now need to end with `.module.css`... rename the files and it should work. Besides that, all the unit tests seem to be passing.

When you fix this, we can merge.

flamecoals:
Now lint is failing... Try `yarn run fix`

web/src/games/cardtable/CribbageBoardB.tsx
cwatsonc: I'll take a look, some of the logic was pretty twisty, I'll review and correct.

PR #743, opened 2021-01-01
Fixes #738

### LINKED ISSUE #738: [Secret Dracula] FBG bar is not visible
Why is there a bug in the first place ? 
Currently, in GameLayout the div rendering the Game is placed in the center of the screen which is good only when game components occupy a fixed size. But in this game, the size of the game board varies as there are special messages appearing at the end. So, I did a trick to keep the top part at the same place -> the trick was to fill the whole screen with a div containing the game board height = 100vh. 

The easiest way to solve this is to increase the z-order of FBG bar, so that it always comes on the top.

flamecoals:
#### Checklist

### DIFF
--- a/web/src/games/secretDracula/board.tsx
+++ b/web/src/games/secretDracula/board.tsx
@@ -126,7 +126,7 @@ export class Board extends React.Component<IBoardProps, IBoardState> {
           ) : null
         }
       >
-        <div style={{ height: '100vh', overflow: 'auto', backgroundColor: 'black' }}>
+        <div style={{ overflow: 'auto', backgroundColor: 'black' }}>
           {/* Leave some space on top for FBG logo and title */}
           <div style={{ height: '12%' }}></div>
 


gk-patel:
well, the reason why 100vh was placed in the first place was to keep the avaters at the same height. 
Now, if messages appear at the bottom, all the avaters move to the top and then they come back down. 
This jumping to top and back to bottom happens everytime - something which is not nice. 

I see this as a temporary solution. If you dont have time to make a permenant fix, I can accept the changes ....
I would prefer a way where we can permenantly fix a div at a particular heigth on the screen.

flamecoals:
Changing the z-index as you proposed did not work. I am not familiar with
the secret dracula implementation to fix this in another way, so it is up
to you to accept this as a temporary solution or not until you or someone
else more familiar has time...

gk-patel (approved):

PR #745, opened 2021-01-02
Feature/cardtable addresses defect - foreign clicks causing state changes

cwatsonc:
#### Checklist

PR addresses clicks on non-active containers causing state changes in active target.  UX now limits the containers active per stage value in onClick. Added test to validate cross click failure.


web/src/games/cardtable/CardContainer.tsx
```diff
@@ -15,6 +16,7 @@ interface ICardContainerProps {
       handleRotateTurn: () => void;
     };
     gameState: {
+      playerID?: string;
       stage?: string;
```
flamecoals, line 20: You could replace the string with a enum, so you get compiler type safety instead of comparing strings.

flamecoals: Same thing for phase

web/src/games/cardtable/CardContainer.tsx
```diff
@@ -47,10 +52,39 @@ const CardContainer: FunctionComponent<ICardContainerProps> = (props: ICardConta
...
 
-  const handleClick = (evt: React.MouseEvent, idx?: number) => {
+  const isClickAllowed = (parentContainer: string): boolean => {
+    if (stage && parentContainer) {
+      switch (stage) {
+        case 'cuttingForDeal': {
+          return name === 'Deck'; //&& currentPlayer === eventClicker;
+        }
+        case 'thePlay': {
+          return (name === 'North Hand' && playerID === '0') || (name === 'South Hand' && playerID === '1');
```
flamecoals, line 62: Can you use an enum here too instead of a string? Comparing string values isn't type safe and very error prone for typos and refactors

web/src/games/cardtable/deals.ts
```diff
@@ -1,56 +1,166 @@
...
+import img6H from './media/png/6H.png';
+import img5H from './media/png/5H.png';
+import img4H from './media/png/4H.png';
+import img3H from './media/png/3H.png';
+import img2H from './media/png/2H.png';
+import imgGB from './media/png/gray_back.png';
+
+export const deckAssets = {
+  AS: imgAS,
+  '2S': img2S,
```
flamecoals, line 57: You can use an enum for the keys too, so you get type safety:

interface Foo {
   [key: NewEnum]: Bar;
} 
let foo:Foo = {};

flamecoals (changes requested):

web/src/games/cardtable/deals.ts
cwatsonc: so I'm a little green in TS I'll have to look at your example.. I've worked with enums in other languages, just not JS/TS...

web/src/games/cardtable/CardContainer.tsx
cwatsonc: addressed in upcoming commit -- added playerEnum, stageEnum, phaseEnum, cardEnum

web/src/games/cardtable/CardContainer.tsx
cwatsonc: added enum logic in next commit.

web/src/games/cardtable/deals.ts
cwatsonc: not sure if I got this completely right, it is addressed in forthcoming commit.

cwatsonc:
@flamecoals wip has been updated with enums where strings were previously used.  The key method you describe didn't work for me, although not sure of the correct behavior -- the docs I encountered were clear that "reverse mapping" with string enumerations isn't supported.

flamecoals (approved):
This looks much better! Thanks! If you have time later to refactor, this could be an enum too:

``
name === 'North Hand'
```

PR #748, opened 2021-01-04
Bingo client side selection

gk-patel:
Last weekend I played Bingo with 25 ppl, and what it was a mess. 
My guess is that when everyone is trying to change G simultaneously, they are getting INVALID modes. 

I have changed the number selection to be local! So now request will only be sent when the user claims bingo. 
Eventhough the data is stored in state, I allow the user to reload the page if needed for whatever reason. 

Before accepting the PR, please tell me if this.props.gameArgs.matchCode is always unique or not ? 
I expected the value of matchCode to equal the URL but it is not. I hope atleast the value is always unique for every game ....


web/src/games/bingo/utils.ts
```diff
@@ -17,3 +17,22 @@ export function inferActivePlayers(G: IGameState, playerID: string) {
   const shoutCallPlayers = Object.keys(G.players).filter((p) => G.players[p].shoutCount > 0 && p !== playerID);
   return [playerID, ...shoutCallPlayers].slice(0, MAX_BACKOFF_CANDIDATES);
 }
+
+export function gameLocalStore(action: string, matchCode: string, key: string, value: any = null) {
+  if (typeof window !== 'undefined') {
```
flamecoals, line 22: You can replace this with a guard statement:
```
if (typeof window === 'undefined') {
  return;
}
```

This way, you have less nesting which makes the code easier to read.

web/src/games/bingo/utils.ts
```diff
@@ -17,3 +17,22 @@ export function inferActivePlayers(G: IGameState, playerID: string) {
   const shoutCallPlayers = Object.keys(G.players).filter((p) => G.players[p].shoutCount > 0 && p !== playerID);
   return [playerID, ...shoutCallPlayers].slice(0, MAX_BACKOFF_CANDIDATES);
 }
+
+export function gameLocalStore(action: string, matchCode: string, key: string, value: any = null) {
```
flamecoals, line 21: Passing a string on `action` to define the behavior of the function is a bad idea, given that the compiler cant validate it is a valid value and typos might lead to unexpected behavior.

I recommend using two functions, one for setting and another for getting in this case.

flamecoals (changes requested):

web/src/games/bingo/utils.ts
```diff
@@ -17,3 +17,22 @@ export function inferActivePlayers(G: IGameState, playerID: string) {
...
+  if (typeof window !== 'undefined') {
+    const storeKey = 'bingoLocalStore';
+    let storeValue = JSON.parse(localStorage.getItem(storeKey)) || {};
+    if (storeValue.matchCode !== matchCode) {
```
flamecoals, line 25: Use session storage instead:
https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

This will work better even if the user has two tabs open.

flamecoals: and you wont need to check for the matchCode at all

web/src/games/bingo/utils.ts
gk-patel: done in new commit.

web/src/games/bingo/utils.ts
gk-patel: Changed to session storage, but will still need match code. 
Otherwise if you start a new game in same window, you will see previous selection.

web/src/games/bingo/utils.ts
gk-patel: done in new commit

flamecoals (approved):

PR #749, opened 2021-01-05
Fix Mergers header

lfkellogg:
Hiding the chat box makes the header disappear. This fixes that by changing the padding to margin, and adjusts the height and overflow-y of the container since the scroll bar is now unnecessary on the board component.

Before:
image

After:
image]

#### Checklist


flamecoals (approved):

flamecoals:
That's great, thank you !

PR #751, opened 2021-01-10
fix #737 avoid duplicated hints for Bulls and Cows

### LINKED ISSUE #737: Bulls and cows hints seem wrong.
**Describe the bug**
It seems that the hints are generated double , see image. 

**To Reproduce**
1. Find a colour which is present on the board 
2. Guess this colour 4 times, it will give you correct/hints for all four.

**Expected behavior**
If a colour only exists once in the solution, I only expect to see one hint if I guess this colour twice.

**Screenshots**

leocaseiro:
#### Checklist


flamecoals:
Great, thank you!

flamecoals (approved):

PR #752, opened 2021-01-17
Full Customization

flamecoals:
#### Checklist


web/src/infra/settings/CustomizationBar.tsx
```diff
@@ -0,0 +1,161 @@
...
+        </IconButton>
+        <Typography variant="h6" component="span" className={css.DialogTitle}>
+          Customization
+        </Typography>
+      </div>
+    );
+  }
+
+  toggleFullCustomizationDialog = () => {
+    this.setState({ showCustomizationDialog: !this.state.showCustomizationDialog });
```
jasonharrison, line 141: Careful with this... `this.setState()` does not immediately mutate state.  It instead creates a pending state transition.  Updating the state like this may cause flaky and hard-to-debug issues.

When the value of the new state depends on the value of the current state, I recommend you enqueue an atomic update, like this:

```typescript
this.setState((prevState => {
  return {showCustomizationDialog: !prevState.showCustomizationDialog};
}));
```

web/src/infra/settings/CustomizationBar.tsx
```diff
@@ -0,0 +1,161 @@
...
+      mode,
+      currentValue: (this.state.customizationState || {})[mode]?.quick,
+      onChange: this._changeCustomValue(CustomizationType.QUICK),
+    });
+  }
+
+  private hasFullCustomization() {
+    const mode = this.props.info.mode;
+    const custom = this.state.customization;
+    if (!custom || !custom.renderFull) {
```
jasonharrison, line 76:
```typescript
if (!custom?.renderFull) { ...
```

web/src/games/secretcodes/customization.tsx
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const renderPredefinedWordsSelect = (
+  onChange: (state?: FullCustomizationState) => void,
+  state: FullCustomizationState,
+) => {
+  return (
+    <Select
+      value={getPredefinedWordsBucket(state)}
+      displayEmpty={true}
```
jasonharrison, line 72: fwiw, a prop passed without a value automatically defaults to `true`

```typescript
<Select displayEmpty={true}>...
// is the same as
<Select displayEmpty>...
```

jasonharrison (changes requested):

web/src/games/secretcodes/customization.tsx
flamecoals: Done.

web/src/infra/settings/CustomizationBar.tsx
flamecoals: Done.

web/src/infra/settings/CustomizationBar.tsx
flamecoals: Done.

jasonharrison (approved):

PR #753, opened 2021-01-19
Do not depend on Coord's instance methods from serialized board state…

### LINKED ISSUE #750: Checkers Double Jump Crashes Game
**Describe the bug**
In online multiplayer mode, the checkers game crashes every time a user faces a double jump scenario.

**To Reproduce**
1. Set up a multiplayer online checkers game
2. Set up a double jump scenario (spot where one user would have two continuous jumps
3. Perform the first jump

**Expected behavior**
The user should be able to make the double jump without the page crashing

**Desktop (please complete the following information):**
 - Mac OS
 - Chrome

mateusazis:
… in checkers.

Fixes #750.

#### Checklist

### DIFF
--- a/web/src/games/checkers/game.ts
+++ b/web/src/games/checkers/game.ts
@@ -45,11 +45,15 @@ export class Coord implements ICoord {
     const y = Math.floor(position / 8);
     return new Coord(x, y);
   }
+
+  static fromICoord(other: ICoord): Coord {
+    return new Coord(other.x, other.y);
+  }
 }
 
 interface ICheckerPieceWithCoord {
   data: ICheckerPiece;
-  coord: Coord;
+  coord: ICoord;
 }
 
 export interface IMove {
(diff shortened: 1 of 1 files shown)


flamecoals (approved):

flamecoals:
Thank you!

leocaseiro:
Thank you @mateusazis!

flamecoals:
@mateusazis Can we get rid of this class entirely? This is a known issue that classes are not serializable, so I am worried another corner case might bring the same issue as we are calling other methods on the game logic. I missed this during review, sorry.

mateusazis:
We can. But, as you said, this doesn't prevent this bug from coming back. The [boardgame.io docs](https://boardgame.io/documentation/#/?id=state) warn about it but also don't prevent it. Some ideas that came to my mind were:

- *At compile time*: annotate the top-level state interface with a [decorator](https://www.typescriptlang.org/docs/handbook/decorators.html) and make sure the class no methods (not sure if this is doable in Typescript).
- *At test time*: between turns, serialize and de-serialize the game state when running in Local mode. This would have made our unit tests catch the bug.

Ideally, the 2nd approach would go into the boardgame.io lib.

flamecoals:
@mateusazis I am not sure if the first approach is feasible either.

Let's go with the second approach, I think this would also help other developers on the boardgame.io community. We can probably allow for a "strict check" during testing on the boardgame.io Client API. I created  https://github.com/freeboardgames/FreeBoardGames.org/issues/754 to keep track of the work.

flamecoals:
For now, I sent you @mateusazis https://github.com/freeboardgames/FreeBoardGames.org/pull/755 to remove the class from checkers entirely.

PR #736, opened 2020-12-29
Introduce "Coord" class in Checkers to group together related methods.

mateusazis:
Giving these functions a more object-oriented face.

#### Checklist

* [X] Use a separate branch in your local repo (not `master`).
* [X] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).

PR #755, opened 2021-01-21
Remove classes from checkers to avoid serialziation bugs

flamecoals:
#### Checklist

### DIFF
--- a/web/src/games/checkers/coord.ts
+++ b/web/src/games/checkers/coord.ts
@@ -3,7 +3,7 @@ export interface ICoord {
   readonly y: number;
 }
 
-export function coord(x: number, y: number): ICoord {
+export function createCoord(x: number, y: number): ICoord {
   return { x, y };
 }
 
@@ -31,8 +31,4 @@ export function fromPosition(position: number): ICoord {
   const x = position % 8;
   const y = Math.floor(position / 8);
   return { x, y };
-}
-
-export function fromICoord(other: ICoord): ICoord {
-  return coord(other.x, other.y);
 }
--- a/web/src/games/checkers/coord.ts
+++ b/web/src/games/checkers/coord.ts
@@ -1,0 +1,38 @@
+export interface ICoord {
+  readonly x: number;
+  readonly y: number;
+}
+
+export function coord(x: number, y: number): ICoord {
+  return { x, y };
+}
+
+export function sum(a: ICoord, b: ICoord): ICoord {
+  return { x: a.x + b.x, y: a.y + b.y };
+}
+
+export function multiply(coord: ICoord, multiplier: number): ICoord {
+  return { x: coord.x * multiplier, y: coord.y * multiplier };
+}
+
+export function toIndex(coord: ICoord) {
+  return coord.x + coord.y * 8;
+}
+
+export function inBounds(coord: ICoord) {
+  return coord.x >= 0 && coord.x < 8 && coord.y >= 0 && coord.y < 8;
+}
+
+export function equals(a: ICoord, b: ICoord): boolean {
+  return a.x === b.x && a.y === b.y;
+}
+
+export function fromPosition(position: number): ICoord {
+  const x = position % 8;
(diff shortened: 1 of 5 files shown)


mateusazis:
Do you prefer instead to just rollback #736 ?

flamecoals:
I think this is better because we have more tests now...

web/src/games/checkers/coord.ts
```diff
@@ -0,0 +1,38 @@
...
+  return a.x === b.x && a.y === b.y;
+}
+
+export function fromPosition(position: number): ICoord {
+  const x = position % 8;
+  const y = Math.floor(position / 8);
+  return { x, y };
+}
+
+export function fromICoord(other: ICoord): ICoord {
```
mateusazis, line 36: This shouldn't be necessary anymore. It was used to convert ICoord -> Coord to use the instance methods but those are now gone.

web/src/games/checkers/game.ts
```diff
@@ -82,18 +39,18 @@ export const INITIAL_BOARD: Piece[] = [
...
 
-const KING_DIRS: Coord[] = [new Coord(-1, 1), new Coord(1, 1), new Coord(-1, -1), new Coord(1, -1)];
+const KING_DIRS: ICoord[] = [coord(-1, 1), coord(1, 1), coord(-1, -1), coord(1, -1)];
 
 export function checkPosition(
   G: IG,
   playerID: string,
   piece: ICheckerPiece,
-  coord: Coord,
+  coord: ICoord,
```
mateusazis, line 53: This name clashes with the "coord" function. Maybe rename the later to "createCoord"?

web/src/games/checkers/game.ts
```diff
@@ -170,10 +127,10 @@ export function getValidMoves(G: IG, playerID: string, jumping?: ICheckerPieceWi
 }
 
 export function move(G: IG, ctx: Ctx, from: ICoord, to: ICoord): IG | string {
-  const fromCoord = new Coord(from.x, from.y);
-  const toCoord = new Coord(to.x, to.y);
-  const indexFrom = fromCoord.toIndex();
-  const indexTo = toCoord.toIndex();
+  const fromCoord = coord(from.x, from.y);
+  const toCoord = coord(to.x, to.y);
```
mateusazis, line 131: These 2 initial conversions should also be unnecessary (see my previous comment on "fromICoord").

web/src/games/checkers/coord.ts
mateusazis, line 36: This shouldn't be necessary anymore. It was used to convert ICoord -> Coord to use the instance methods but those are now gone.

web/src/games/checkers/game.ts
flamecoals: Done.

web/src/games/checkers/game.ts
flamecoals: Done.

web/src/games/checkers/coord.ts
flamecoals: Done.

web/src/games/checkers/game.ts
```diff
@@ -127,8 +127,8 @@ export function getValidMoves(G: IG, playerID: string, jumping?: ICheckerPieceWi
 }
 
 export function move(G: IG, ctx: Ctx, from: ICoord, to: ICoord): IG | string {
-  const fromCoord = coord(from.x, from.y);
-  const toCoord = coord(to.x, to.y);
+  const fromCoord = createCoord(from.x, from.y);
```
mateusazis, line 130: Sorry, what I meat here is that you don't need the "fromCoord"and "toCoord" variables at all. You can just pass "from" and "to" directly to "toIndex" below.

mateusazis (approved):

web/src/games/checkers/game.ts
flamecoals: Done

Issue #763, opened 2021-03-01
[Bug] Bombs and Bunnys  - Game enters state with no legal moves


Spooky-0:
**Describe the bug**
I played Bombs and Bunnies a while back and we got into a state with no legal moves. I am sorry to say that I don't really remember the exact details.

**To Reproduce**
We played a six player game, and then the second player died (ran out of cards) there was no player with an active move. The game was stuck.

**Expected behavior**
That the game would continue

flamecoals:
Thanks for reporting this @Spooky-0. @b-hub do you have any idea what might be? It would be great if we could reproduce the issue

b-hub:
Not sure, but I will add some 6 player tests and see if anything turns up

b-hub:
I think this issue can be closed now @flamecoals

flamecoals:
Thank!

PR #767, opened 2021-03-11
B hub skulls

b-hub:
#763 Should hopefully resolve the issue raised by @Spooky-0

The bug was caused by a custom play order which removes knocked out players. However, when calculating the first() player during the initial_placement phase the order position returned used the old play order instead of the new custom play order. This meant in some cases the order position was out of range and ended up causing ctx.currentPlayer = undefined.


Spooky-0:
Sweet! Glad to hear that you found (and fixed 👏) the bug!

b-hub:
Thanks for your report @Spooky-0! Hopefully you can get a full game of 6 in now :)

Spooky-0:
Btw one more issue - but I'm not sure if it is actually an issue: once when playing I had to turn cards up. But exactly then I had some internet issue. So I was able to turn over cards and see their value - bomb orbunny. I refreshed my browser and it was like nothing had happened - the cards were turned back. 
This is obviously a problem since players could easily cheat. 
Did you use the Hiden Game state machanic from boardgames.io? If not, it's very easy to implement. The only thing that might be a bit weird ist the game tests since there the tester (player1 for example) might not have access to the values.

Let me know if this is indeed the case and we can work on fixing it together

flamecoals:
Great, thanks @b-hub !

flamecoals (approved):

PR #772, opened 2021-04-18
Localization with next-i18next

richardaum:
#### Checklist

Discussion: https://github.com/freeboardgames/FreeBoardGames.org/issues/633

### Missing work
* [ ] Automated tests (suggestions accepted!)

### Summary

Everything here is **discussable**. This is my **first PR** and I'm very happy to help with this project, so feel free to comment about things that should be made differently. 

Also, I'm not sure about what would be a small PR - so I decided to follow by which would be the minimal piece of functional code, which describes the intention. If even so, it got large I could break it up, no problem. A personal tip to review this one would be to follow the commits since they express exactly what I'm describing here.

<hr/>

**chore: add next-i18next dependency**

Adding the `^7.0.1` version because that version is compatible with our Next.js version, which is `9.x`.

<hr/>

**refactor: apply composition pattern**

As discussed previously, we have multiple Class components and, right now, we are not changing them to Functional components - that means we need the `withTranslation` HOC to have access to the translations (instead of the `useTranslation` hook).

Including one more HOC to some components would like to this kind of implementation:

```typescript
export default const enhance = connect(mapStateToProps)(withTranslation('namespace')(GameModePicker));
```

To prevent that, I've added the [`recompose`](https://github.com/acdlite/recompose) library, which does a pretty much good job, allowing composability over HOCs in a functional way. We also get for free the opportunity to make other developers know which props are internal (injected by some HOC) and external (provided directly by usage):

```typescript
const enhance = compose<IGameModePickerInnerProps, IGameModePickerOutterProps>(
    connect(mapStateToProps), 
    withTranslation('GameModePicker'),
);

export default enhance(GameModePicker);
```

_Mental note: Maybe this addition is too much for this PR's context._

<hr/>

**chore: add next and i18next configs**
Just followed the README.me from the `next-i18next` library (https://github.com/isaachinman/next-i18next/tree/v7.0.1), which guided me to add these files.

<hr/>

**chore: add i18next-hmr dependency**
Reference: https://github.com/isaachinman/next-i18next/issues/626

The library's author has well described that the developer experience may not be a focus right now, so live reloading changed JSON files is not achievable by his library. But, there is another library `i18next-hmr` which achieves pretty much that! And it only runs for `development` NODE_ENV.

<hr/>

**feat: add translation for single text**

Last but not least, I've randomly picked a piece of a React component to "translate" to English. `GameModePicker` was the chosen one.

I need to make some assumptions about the usage, they are:

- Following the i18next recommendation, all JSON files live on the `public/static` folder. I've tried to move them to live colocated with the TSX file, but due to the way i18next works on Next.js - the client also being able to make a request and get the translation async - I decided to keep using this way.
- There is a `common.json` file to have common strings we use. It could also be split for more granularity, but I prefer each component having its own namespace, making it easy to correlate the component and its messages, and having a **single source of truth per component** about messages, which is important for projects which could grow up.
- As mentioned on docs, in order to use `withTranslation` you must provide which namespace you are consuming. 
- Also each Next.js page contains a `namespacesRequired` property, which tells the server which namespace it should preload to speed up things. Everytime we create a new namespace, we need to add it into that property for each page it is used, otherwise the lib will show a warning.
 
<hr />

**test: mock `t` function**
also: _make only `t` function to be available_
_Not sure_ if we should test the translations. Following that, `t` must be mocked since it comes from the translation provider.

<hr />


vdfdev:
This is looking great, I will look deeper into it tonight!

web/src/infra/gameInfo/GameInfo.tsx
```diff
@@ -89,7 +89,10 @@ class GameInfo extends React.Component<gameInfoProps, {}> {
     if (!GAMES_MAP[gameCode] && ctx.res) {
       return generatePageError(404);
     }
-    return { gameCode };
+    return {
+      gameCode,
+      namespacesRequired: ['common', 'GameModePicker'],
```
vdfdev, line 94: Isn't this supposed to be "GameInfo" given that this is another page?  Was this changed intended ?

vdfdev:
Some points:

- Regarding using recomponse library: This is a great idea, I did not know about it! Thanks for the improvement :)
- Regarding the namespace by component and unified place for all translations: LGTM! This is great because it will help translators to know where each string goes. I agree that having some repetition is OK, and having a single place for the "common" strings (i.e. "Play") is great too!
- Regarding tests: Because this is just a function call and there is no control flow, I don't think we need to test that the library is doing what is supposed to do.


Overall, this PR LGTM :). I will just double check it works and merge it.

web/src/infra/gameInfo/GameInfo.tsx
richardaum: I thought we needed to specify each namespace required by inner components, but I need to read the docs again to see if it's really necessary. Can I do that tomorrow? :)

vdfdev: I am reverting these changes before merging

vdfdev: Nevermind! This is WAI, I got this error when I removed:
```
You have not declared a namespacesRequired array on your page-level component: withRouter(GameInfo). This will cause all namespaces to be sent down to the client, possibly negatively impacting the performance of your app. For more info, see: https://github.com/isaachinman/next-i18next#4-declaring-namespace-dependencies
```

vdfdev (approved):

web/src/infra/gameInfo/GameInfo.tsx
vdfdev: You were right :), we need to specify the namespace

richardaum: Yeah! I don't like to have to specify it twice, but it takes advantage of SSR to speed up things. 

https://github.com/isaachinman/next-i18next/tree/v7.0.1#4-declaring-namespace-dependencies

> The `withTranslation` HOC is responsible for passing the `t` function to your component. It enables all the translation functionality provided by `i18next`. Further, it asserts your component gets re-rendered on language change or changes to the translation catalog itself (loaded translations). More info can be found [here](https://react.i18next.com/latest/withtranslation-hoc).

> By default, `next-i18next` will send _all your namespaces_ down to the client on each initial request. This can be an appropriate approach for smaller apps with less content, but a lot of apps will benefit from splitting namespaces based on route.

> To do that, you need to return a `namespacesRequired` array via `getInitialProps` on your page-level component. You can see this approach in [examples/simple/pages/index.js](./examples/simple/pages/index.js).

> Note: `withTranslation` provides namespaces to the component that it wraps. However, `namespacesRequired` provides the total available namespaces to the entire React tree and belongs on the page level. Both are required (although you can use `Trans` instead of `withTranslation` if desired).

PR #780, opened 2021-04-22
Localized href/route using a Link component

richardaum:
#### Checklist

### Goal

Main goal here to support translated URLs like `/pt/jogar/xadrez`, `/en/play/chess`, etc.

#### Discussion

https://github.com/freeboardgames/FreeBoardGames.org/issues/633#issuecomment-823716411

#### Summary

**feat: add pt translation for GameModePicker text**

Simply add translations to another language (pt) to have visual feedback when changing between languages.

**refactor: use values from next-i18next-config.js**

Minor refactory to prevent duplicated config values between `infra/i18n/config` and `next-i18next.config` files.

**chore: enable localeSubpaths config**

`localeSubpaths` enables using a locale as prefix (e.g. /pt/play/bingo or /en/play/bingo). This is partially what we desire in terms of URL translation. Unfortanelly, `next-i18next` does not support omitting the locale prefix and using the `defaultLanguage` value. See discussion here https://github.com/isaachinman/next-i18next/issues/472. :disappointed: 

Additionally, we needed to add `rewrites` and `redirects` to properly map the URL translated and try to redirect to a valid URL in case something gots weird (e.g. `/pt/play` instead of `/pt/jogar`).

**feat: add Link component to use a localized href**

Inspired by the approach used by `next-i18next` where it wraps the Link component to properly prefix the `href` provided, there is also a Link wrapper on our side. This wrapper is responsible to translate all routes. A React hook was created to separate this responsibility. Maybe we need extra work later to do the same thing for dynamic navigation (i.e. Router.push).

There is a `routing.json` file, which commands all route translations. Every time a new page is translated, it should be added there. To maintain consistency between path params over the languages, there is a unit test that checks that file.

**chore: enable prettier to work over locale files**

The previous definition to apply prettier on JSON locale files wasn't working properly. After digging a bit more, the change proposed here now works fine.


vdfdev:
Cool! I will take a look into it tonight :)

vdfdev:
Wow, great job here! I have only minor improvements to the code, I will push them directly as I think this will get us done faster :)

vdfdev:
Some few improvements I made on the code:
- Removed the try block. The case where the params on the old url do not match with the params on the new url is already covered by the unit test you wrote, so it should never happen. Try blocks can hide other bugs, so it is better to avoid.
- Reduced nesting by using guard statements: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html
- Renamed "default" to "original" on the routes json, as .default seems to be a JS module 
- Added some test cases

This is an impactful change, as it will change most of our URLs to append /en/ in the beginning (i.e. freeboardgames.org/play/chess -> freeboardgames.org/en/play/chess). I think that is OK because we are setting the permanent redirects, so I hope it does not impact the traffic sent by search engines (but there is a bit of risk).

However, before we merge, I want to test it a bit more thoroughly to make sure nothing breaks. For instance, I noticed some issues we need to fix beforehand:
- [ ] The header logo on the website is not using Link, so it is not appending /en/, and somehow it is not getting redirected either.
- [ ] The about link on the bottom is not using Link. I dont have a strong opinion if we should translate this URL or not, I am more worried about the /play/ ones.
- [ ] The /docs seems to be using Link, and because it is static content, it gives 404 for /en/docs. I don't think we should append the language in that case, should always be /docs as we are not translating documentation.
- [ ] Need to upgrade some e2e tests
- [ ] Need to upgrade our sitemap to use /en/ now...

richardaum:
Thank you for working with also on this one!

Agreed on all you said.

I am also afraid of having `/en` prefix over all urls. `next-i18next` is
forcing that behavior over the entire Next application, so we would need to
override that behavior directly on lib source code, maybe forking and
having our own version - which is cool but hard-working.

What do you think about we trying another translation lib which supports
not prefixing /en, by default?

Then we could fix those issues you pointed.

vdfdev:
I was reading some blogs about these optimizations and as long as we are issuing a 301 permanent redirect and update the sitemap it shouldn't affect much our performance. I think it is a risk that we should take in order to have a more consistent url and so on, and it is worth doing it over forking the lib or scrapping all the work already done and changing the lib... Let's do it :) we just need to figure out these details I mentioned above

richardaum:
> Should we translate all paths? /about, /room/, /match/ 
Maybe we could leave the translation of those to be made in a second moment, with less priority.

> How to deal with translated game codes, game name, game description, etc? 
After you said on discord, now I got what you said. Let me think a little bit more. 

> Can the i18n function work on IGameDef and index.ts directly? 
As far as I could tell, no. https://github.com/isaachinman/next-i18next/issues/983 and https://github.com/isaachinman/next-i18next/issues/1127 point in that direction, but I'd like to dig it first.

> Maybe we should create a new field called "alias" for the URL name that can be translated, while the "code" needs to be constant in all languages (used in database and etc). 
Totally agreed with you! Do you think we could do that on a second PR?

> We might also want to have a field for each game that will list all languages this game is translated to, as I don't think we will be able to translate all games to all languages simultaneously.
Good! Maybe this one would be on a third PR?

richardaum:
> The /docs seems to be using Link, and because it is static content, it gives 404 for /en/docs. I don't think we should append the language in that case, should always be /docs as we are not translating documentation.

Just to be sure, that is the behavior we want. Right?

Btw, `/docs` is using a [Button](https://github.com/freeboardgames/FreeBoardGames.org/blob/96d3e0deb885d3bde539ddd683596e2e2e6bc6ad/web/src/infra/home/Header.tsx#L42) from `@material-ui` - which happens to become a `<a />` after rendered.

richardaum:
Issues:

- [x] The header logo on the website is not using Link, so it is not appending /en/, and somehow it is not getting redirected either.
- [x] The about link on the bottom is not using Link. I dont have a strong opinion if we should translate this URL or not, I am more worried about the /play/ ones.
- [x] The /docs seems to be using Link, and because it is static content, it gives 404 for /en/docs. I don't think we should append the language in that case, should always be /docs as we are not translating documentation.
- [x] Need to upgrade some e2e tests
- [] Need to upgrade our sitemap to use /en/ now...

richardaum:
@vdfdev you mentioned that we "Need to upgrade some e2e tests", right?. do you mean fixes the current ones which ~are~ were failing or something else?

vdfdev:
Definitely can be done on the second PR! Yes, upgrade the e2e tests was just fixing it, thanks! 

Thanks for already getting into this so quick, I will take a look soon !

vdfdev:
I noticed now that the sitemap.xml is not being generated after merging the previous PR. I am looking into it.

vdfdev:
Tested a bit more thoroughly now, found some new issues:

- [] Users are being redirected to /room instead of /en/room after clicking on the new match button or replay button, check home page lobby too
- [] Users are being redirected to /match instead of /en/match after the match starts
- [] Link to invite friens on the room does not have the /en as it should

I can work on this in a few hours after lunch :)

web/server/sitemap/sitemap.ts
```diff
@@ -5,19 +5,13 @@ import { template } from './sitemap.template';
 import { Manifest, Url } from './types';
 
 export const generateSiteMapXML = (options: { manifest: Manifest; staticDir: string; host: string }) => {
-  if (isDevelopment()) return;
```
richardaum: @vdfdev I am adding back this restriction because on `development` there is no `app.pagesManifest`. I don't know why it doesn't exist. But without that guard, the app breaks when running on dev.

vdfdev: Sure! Sorry about that, I should have tested...

richardaum:
@vdfdev I have a suggestion to follow that would allow us to play safe about this PR: I am going to add a feature toggle (using a new environment variable called `I18N_ENABLED`). 

But why?
- This way we can safely work on code without worry about releasing something that could break.
- Prevent us to let a PR too much time opened
- Prevent us to have a giant PR to review (now and in the future). I have noticed a have a lot of code to be pushed and I am trying to follow that rule about small PRs :D

Does that make sense?

vdfdev:
Sure! Sounds good 😁

vdfdev:
Ok! I think we can merge now because with the feature flag it does seem to keep the current behavior in prod.

Can you let me know how to activate this flag? I tried `I18N_ENABLED=true yarn run dev` but it did not work... I know how to set some envs on Docker but not locally...

vdfdev (approved):

vdfdev:
Trying to follow the advice from https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables ... I left a `.env.local` with `I18N_ENABLED=true` and ran `yarn run dev:internal` inside the `web` folder, but no luck... Let me know if you can make it work locally with this flag enabled.

PR #781, opened 2021-04-24
Sitemap generation refactory

richardaum:
#### Checklist

#### Goal

A refactor for the sitemap code, no behavior change is expected.

#### Motivation

Currently, the sitemap generation code is written directly on `server/web.ts`. Moving to a dedicated module, leading to better readability of both `server/web.ts` and also of the sitemap generation which now lives on `sitemap/*`.

Along with this refactory, a unit test is being added. For that, it was required to change `jest.config.js` file because previously it was only getting tests from `src` folder.


vdfdev (approved):

vdfdev:
Looks great!!!

PR #782, opened 2021-04-24
Prevent cyclic dependencies using ESLint

richardaum:
#### Goal

Prevent cyclic dependencies as it silently prevents the server from starting.

#### To do

- [x] This PR has a target that is not the `master` branch. Check this item if the target branch is already merged.

#### How?

Using eslint-plugin-import + eslint-import-resolver-typescript, we enable ESLint to see where are the cyclic dependencies.

**Disclaimer**: 

After installing the plugin, it was required to fix a cyclic dependency located at the `cornerus` game. That cyclic dependency was found on a typing import. In that case, Typescript handles in a way it doesn't hurt the app in runtime, otherwise, it would also break the application. 

But why change them, if the cyclic typing imports don't break the application? For two reasons:
1) Currently the plugin isn't able to distinguish between importing a typing definition or importing a module. https://github.com/benmosher/eslint-plugin-import/issues/1453
2) Given the first reason, it ends up bringing consistency about cyclic dependency, not having it at all.


vdfdev (approved):

vdfdev:
Feel free to merge this whenever you think it is ready, it is approved

richardaum:
Tomorrow I will take a closer look at this PR in order to check before merging.

PR #785, opened 2021-04-27
I18N_ENABLED renamed to NEXT_PUBLIC_I18N_ENABLED

richardaum:
### Goal
Make i18n feature toggle work properly when we desire to enable it.

### Description
You could enable the i18n feature by provided NEXT_PUBLIC_I18N_ENABLED=true to the process any way you prefer (dotenv, inline on terminal command, etc.)

**refactor: change variable to NEXT_PUBLIC_I18N_ENABLED**
- Next.js prefer that public environment variables have a specific prefix in order to have its **value** exposed to the client-side.  (see https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables)

chore: by default env vars are forced to string …
- It needs to a be string because Node.js force that to be one, even if we assign a boolean (see https://nodejs.org/api/process.html#process_process_env)

fix: localeSubpaths and i18n should be provided as objects to be merged
When using spread over a boolean conditional, we need to make sure the second part of the statement is an object we want to merge, otherwise, it will merge only the value which is wrong in this case.


vdfdev:
It worked! :D I just had to add `NEXT_PUBLIC_I18N_ENABLED=true` to `.env.local` :). Great job :)

vdfdev (approved):

PR #786, opened 2021-04-28
[i18n] apply the i18n redirect logic (part. 2)

richardaum:
### Goal

This PR is part of a major change of making sure every place that needs changing the application route is using `infra/i18n` instead of `next/link` or `next/router`.

The main goal is to change the implementation of programmatic redirects, especially for Play Again on AI, LocalFriend, and otherwise.

### Summary

**refactor: extracting `translateHref` to a pure function**

In order to reuse the route/href translation implementation on `withRouter` and `useRouter`, that function is being extracted to an isolated and reusable pure function.
<hr />

**feat: wrap Router using translateHref**
**refactor: organize i18n by layer**

The `translateHref` function is being used on `push` and `replace` to wrap the router behavior and promote route/href translation. Using layers (components, hocs, hooks...) to promote a consistent organization.
<hr />

**refactor: moving components to i18n folder**

Since `Link` is related to i18n context, it looks better to move it to the `i81n` folder.
<hr />

**feat: add useRouter hook**
**feat: add withRouter HOC**

Adding elements to support migration to localized href/route - those elements are wrapping `translateHref` function.
<hr />

**refactor: expose Router as NextRouter**

Just to prevent import `next/router` directly, making `infra/i18n` act like an abstraction.
<hr />

**test: prevent i18n warnings when running tests**

I've noticed after adding i18n it adds a annoying warning on test, because there is no I18nProvider. To prevent that warning, I'm mocking the I18nContext only for test scope.
<hr />

**test: simplify Link.test**

Rather than mocking language, use i18n.changeLanguage function.
<hr />

**i18n: add withRouter hoc to GameDarkSublayout**

Changing GameDarkSublayout is part of the goal here. As `withRouter` is only being used to get `query.matchId`, there is nothing special here.
<hr />

**refactor: use mocked-env to reduce code when mocking process.env**

<hr />

**feat: change redirect for programmatically scenarios for Play Again on AI, LocalFriend, and otherwise**

The goal is to change PlayAgain behavior to redirect to translated URLs.
<hr />


vdfdev:
Looking great again! Tomorrow I have some time off, so I can help you merging this!

richardaum:
Thanks :D
No problem at all. I'm still working here to fix some tests but having some
hard time on mocking stuff on jest. Almost finishing.

vdfdev:
LGTM! The play again is now indeed going to the translated URL, just tested, and the code looks great! Thanks for thoroughly testing the code with unit tests

vdfdev (approved):

PR #789, opened 2021-05-01
[i18n] change next/router for infra/i18n (part. 5) 

richardaum:
### Goals

- Replace next/router for infra/i18n for some redirects on /play, /room, /g.
- Replace next/router for infra/i18n for SEO Breadcrumbs (because of pathname).
- Replace onRouteChangeComplete for a subscription of a listener on a specific event: `routeChangeComplete` (see https://stackoverflow.com/a/53933774).
- Translate shareable room link by making `translateHref` function also call `lngPathCorrector` - which is an internal function from next-i18next.

These changes should finish all required changes to make navigation compliant to i18n :tada:

### DIFF
--- a/web/src/infra/game/AIOrLocalGame.tsx
+++ b/web/src/infra/game/AIOrLocalGame.tsx
@@ -17,7 +17,7 @@ interface AILocalGameProps {
   mode: string;
 }
 
-export default class extends React.Component<AILocalGameProps, {}> {
+export default class AILocalGame extends React.Component<AILocalGameProps, {}> {
   render() {
     if (this.props.gameDef) {
       return (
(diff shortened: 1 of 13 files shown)


vdfdev:
Interesting enough, `translateHref` does not seem to be adding the /en/ on the room page:

image

vdfdev:
So the clipboard still misses /en/
image

I can try to help you debugging it tomorrow... I am going to sleep now :P

richardaum:
I think now it is fine... Last change I made I introduced a minor bug. I think it is now fixed.

Captura de tela de 2021-05-01 12-01-45

vdfdev:
It worked and looks great! Merging now, and then I will try to flip the flag in staging and then prod.

vdfdev (approved):

PR #792, opened 2021-05-02
[i18n] Game translation structure

richardaum:
### Goals

Propose a game translation structure, which provides a way to translate the game definition (name, description, text instructions, and video id instructions). The game code is not being translated here yet.

#### Proposal

- Each game has a file named with its code (e.g. `public/static/locales/en/chess.json`) - each language needs a file like this one.
- In case of a game does not have a translating file, it will fallback for the content located where the game is defined (e.g. `src/games/chess/index.ts`).
- `withTranslate` is a decorator which knows how to translate string which are specific for a game (not only its definitions). The game is obtained from pages where `gameCode` is a path param. Also, on those pages and inner components, you could use `useCurrentGame` to have access to the `gameDef` instance.
- Inside a component:
  - if you need to translate something related to the infra (e.g. Play label) you should use the `t` function which is injected by `withTranslation` HOC.  Extend your component props with `WithTranslation` (or maybe `Pick<WithTranslation, 't'>` type for better typing.
  - But if you need to translate something that is strictly related to the game (e.g. Chess) you should use `withTranslate`, which injects a `translate` function. Extend your component props with `WithTranslate` type for better typing.
  - Why? `withTranslate` is a copy of `withTranslation` which already selects the game namespace for translation (e.g. `public/static/locales/en/chess.json`).
- As you may have noticed, once a game is translated, the `IGameDef` instance won't need name, descriptions, and instructions properties anymore. But right now, these values serve as a fallback for untranslated games.

#### Other stuff

There are other minor changes made here:
- Adding a node script to run `gen:games` from web folder
- Making sure env (NEXT_PUBLIC_I18N_ENABLED) is defined on each test (to prevent side effects if the variable is already defined somewhere else)

### DIFF
--- a/web/src/infra/game/GameProvider.tsx
+++ b/web/src/infra/game/GameProvider.tsx
@@ -1,0 +1,18 @@
+/* eslint-disable react/prop-types */
+import { GAMES_MAP } from 'games';
+import { IGameDef } from 'gamesShared/definitions/game';
+import { createContext, FC, useContext } from 'react';
+import React from 'react';
+
+const Context = createContext<GameContext>({} as GameContext);
+
+export const useCurrentGame = () => useContext(Context);
+
+export const GameProvider: FC<{ gameCode: string }> = ({ children, gameCode }) => {
+  return <Context.Provider value={{ game: GAMES_MAP[gameCode], gameCode }}>{children}</Context.Provider>;
+};
+
+export interface GameContext {
+  game?: IGameDef;
+  gameCode?: string;
+}
--- a/web/src/infra/i18n/hocs/withTranslate.tsx
+++ b/web/src/infra/i18n/hocs/withTranslate.tsx
@@ -1,15 +1,13 @@
 import { getDisplayName } from 'recompose';
 import { nextI18Next } from '../config';
 import React from 'react';
+import { useCurrentGame } from 'infra/game/GameProvider';
 
-export const withNamedT = <TOutterProps, TReturn extends string = string>(
-  codeGetter: (props: TOutterProps) => TReturn,
-  propertyName: string = 'translate',
-  options: { withRef?: boolean } = {},
-) => {
+export const withTranslate = (propertyName: string = 'translate', options: { withRef?: boolean } = {}) => {
   return function Extend(WrappedComponent) {
     function I18nextWithTranslation({ forwardedRef, ...rest }) {
-      const ns = codeGetter(rest as TOutterProps);
+      const { game } = useCurrentGame();
+      const ns = game?.code;
       const [t] = nextI18Next.useTranslation(ns, rest);
 
       const passDownProps = { ...rest, [propertyName]: t };
--- a/web/src/infra/room/GameSharing.test.tsx
+++ b/web/src/infra/room/GameSharing.test.tsx
@@ -3,6 +3,7 @@ import { GameSharingInternal as GameSharing } from './GameSharing';
 import { render, fireEvent, RenderResult, cleanup } from '@testing-library/react';
 import { I18n } from 'next-i18next';
 import { mock } from 'jest-mock-extended';
+import mockedEnv from 'mocked-env';
 require('@testing-library/jest-dom/extend-expect');
 
 const GAME_LINK = 'http://localhost/room/fooroom';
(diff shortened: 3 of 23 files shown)


web/src/infra/game/GameProvider.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { IGameDef } from 'gamesShared/definitions/game';
+import { createContext, FC, useContext } from 'react';
+import React from 'react';
+
+const Context = createContext<GameContext>({} as GameContext);
+
+export const useCurrentGame = () => useContext(Context);
+
+export const GameProvider: FC<{ gameCode: string }> = ({ children, gameCode }) => {
+  return <Context.Provider value={{ game: GAMES_MAP[gameCode], gameCode }}>{children}</Context.Provider>;
```
vdfdev, line 12: This is a great idea, a great way for us to remove the global references to GAMES_MAP currently spread out all over the code base.

web/src/infra/i18n/hocs/withTranslate.tsx
```diff
@@ -0,0 +1,35 @@
+import { getDisplayName } from 'recompose';
+import { nextI18Next } from '../config';
+import React from 'react';
+import { useCurrentGame } from 'infra/game/GameProvider';
+
+export const withTranslate = (propertyName: string = 'translate', options: { withRef?: boolean } = {}) => {
+  return function Extend(WrappedComponent) {
+    function I18nextWithTranslation({ forwardedRef, ...rest }) {
+      const { game } = useCurrentGame();
+      const ns = game?.code;
```
vdfdev, line 10: NIT I think it would be more readable if you use more descriptive variables, i.e. "namespace" instead of "ns".

richardaum: Great suggestion!

web/src/infra/i18n/hocs/withTranslate.tsx
```diff
@@ -0,0 +1,35 @@
...
+      const [t] = nextI18Next.useTranslation(ns, rest);
```
vdfdev, line 11: This seems great, by using I18Next `useTranslation` it seems like you will be avoiding loading other langauges files, right? So the user seeing the page in portuguese wouldn't have to download the description of the game in french, for instance :)

web/src/infra/room/GameSharing.test.tsx
```diff
@@ -45,12 +46,25 @@ describe('GameSharing', () => {
...
     jest.useFakeTimers();
     const button = wrapper.getByLabelText('Copy');
     fireEvent.click(button);
     expect(wrapper.getByText('Copied!')).toBeInTheDocument();
     jest.runAllTimers();
     expect((window as any).copyClipboardMock).toHaveBeenCalledWith(GAME_LINK);
+    restore();
+  });
+
+  it('should copy link when i18n is enabled', () => {
```
vdfdev, line 60: Wouldn't it be more simple if we get rid of this flag? It is already active in production so I think we don't need it anymore.

vdfdev (approved):

vdfdev:
I love the simplicity of this solution for the game developers! Great job, again :)

image

I approved, feel free to merge whenever you feel ready.

web/src/infra/i18n/hocs/withTranslate.tsx
richardaum: Yeah! `useTranslation(ns, ...)` loads a namespace with name = `game.code`, but only for the current language.

web/src/infra/room/GameSharing.test.tsx
richardaum: Could we wait a bit longer? I'd like to be able to turn off the feature at any moment, at least until we have a game fully translated. :D

vdfdev:
Some questions:

- We cannot move the languages files inside the game folder, right? It needs to be on `public/static/locales` because it is a i18next namespace... It would be a better developer experience if we could get everything on the same folder, but in lieu of that we can write documentation on how to translate their games.
- I think we should create a new field on the gameDef, called "urlAlias", and use this instead of the "code" to determine the URL. This way we can translate the "urlAlias" for each language, and have links like `/pt/jogar/xadrez`, and we will need a mapping between urlAlias -> gameCode to be able to find the i18n namespace. Do you think this solution would work?
- How can we get the translated properties in pages that dont have the game code? For instance, on the home page it is still showing "Chess" instead of "Xadrez"...
image
- Can we somehow not show the game if it is not translated on the home page for a given language? Or at least show an warning asking for contributions...

web/src/infra/room/GameSharing.test.tsx
vdfdev: Sure, sounds good :)

richardaum:
> It would be a better developer experience if we could get everything in the same folder.

Definitely! What about we discuss how can we make it possible somehow?

> I think we should create a new field on the gameDef, called "urlAlias", and use this instead of the "code" to determine the URL.

Yeah. I am working on the next PR that will enable that behavior. The solution looks similar except that I have created a field called `codes`:
```
codes?: Set<string>;
````

> How can we get the translated properties in pages that don't have the game code?

It would be made by using the game code from the array used to render that list. Then we use that raw `t` function using the game code as the namespace.

> Can we somehow not show the game if it is not translated on the home page for a given language? 

Do you mean to hide the game from the list? Yeah, we just need to work on something to flag/ban that game from a specific language.

vdfdev:
Sounds good!

vdfdev:
One way we can get the "everything in the same folder" goal, is to make the `gen:games` script copy the files over. should be simple.

PR #793, opened 2021-05-02
chore: convert next.config to typescript

richardaum:
### Goal

Allow better code reuse between server and web folders, by enabling next.config.js to be written in Typescript.

By default, the framework does not allow that. Then I took advantage of the fact we already have a custom server, which is already being transpiled to Typescript. The only thing remaining was importing the config and update the import statements.


vdfdev:
We might need to update the Dockerfile too..

Which reminds me on something on my TODO list, I need to make our e2e be really e2e and spin a minikube cluster... I will do that next after I finish the horizontally scaling project

vdfdev:
And it seems like we are getting an OOM issue on the `next build` ?

richardaum:
> And it seems like we are getting an OOM issue on the `next build` ?

Oh! Thanks for warning me! I am going to better investigate by tomorrow about the OOM.

richardaum:
> We might need to update the Dockerfile too..
> 
> Which reminds me on something on my TODO list, I need to make our e2e be really e2e and spin a minikube cluster... I will do that next after I finish the horizontally scaling project

We don't need to copy next.config.js or next-i18next.config.js anymore. Because now next.config.ts is imported explicitly and webpack is now able to bundle it together with the other files. So the update would be just removing it from Dockerfile.

vdfdev:
I think it did not work... By running `./docker-build.sh` I got this error:

``
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
The command '/bin/sh -c yarn run build' returned a non-zero code: 1
Sending build context to Docker daemon  2.668MB
``

The script though doesn't exist and still creates the image, but the image is the same as the previous one. I think we need to change the script to abort in case it fails.

vdfdev:
https://github.com/freeboardgames/FreeBoardGames.org/pull/795 fixes the script issue I mentioned above.

richardaum:
I have found the issue by running `yarn build:storybook` which was indicated the culprit: a missing plugin o `.storybook/webpack.config.js`.  I am running `./docker-build.sh` again to check if there are more errors.

The weird thing here is that `./docker-build.sh` was swallowing the error stack trace while running the command directly on my computer is fine.

vdfdev (approved):

PR #807, opened 2021-05-12
[i18n] Translate Home page and its components

richardaum:
### Goal

Translate Home page and its components

Preview

image


vdfdev:
Wow, it is amazing to see this in portuguese :) Muito bacana

vdfdev (approved):

PR #810, opened 2021-05-13
[i18n] Replace raw english text: chess

vdfdev:
Reverts freeboardgames/FreeBoardGames.org#809


vdfdev:
@richardaum FYI, I messed up the order of the merging, so I had to revert the reversion :P

vdfdev:
I am having issues merging this, the CI is failing. It got late for me already, so I will keep looking into it tomorrow

web/src/games/chess/board.test.tsx
```diff
@@ -67,7 +52,7 @@ test('render board - all states - local friend', () => {
       currentPlayerMoves: 0,
     },
   });
-  expect(board.html()).toContain('black won');
+  expect(board.find('[data-testid="gameOverText"]')).toContainText('black_won');
```
richardaum, line 55: These strings with `_` aren't expected here. It should be `black won` because that is the translation.

PR #822, opened 2021-05-13
[i18n] Replace raw English texts on infra/pages components: ErrorPage

richardaum:
- [x] feat: replace raw English texts on infra/pages components: ErrorPage

### DIFF
--- a/web/cypress/integration/mobile/MobileView.ts
+++ b/web/cypress/integration/mobile/MobileView.ts
@@ -1,5 +1,5 @@
 it('visits GameInfo page on mobile', () => {
-  cy.visit('/play/takesix');
+  cy.visit('/en/play/takesix');
   cy.get('[data-testid="MobileViewDiv"]').should('exist');
 });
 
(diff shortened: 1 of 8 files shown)


vdfdev:
This is the error we are getting:

```

     CypressError: Timed out retrying: `cy.click()` failed because this element is detached from the DOM.

`<div class="jss10 jss28" style="background-image:url(/_next/static/images/thumbnail-500-d9aabb6f904731e30bde338f584428ff.jpg)" data-testid="gamecard-tictactoe">...</div>`

Cypress requires elements be attached in the DOM to interact with them.

The previous command that ran was:

  > `cy.get()`

This DOM element likely became detached somewhere between the previous and current command.

Common situations why this happens:
  - Your JS framework re-rendered asynchronously
  - Your app code reacted to an event firing and removed the element

You typically need to re-query for the element or add 'guards' which delay Cypress from running new commands.
```

vdfdev:
I think what is happening is a race condition between the redirect from / to /en and the click. I think changing it to go directly to /en will fix it. I am testing now the fix.

vdfdev:
That did not fix it for me. It still happening 100% of the time when I run it locally. Maybe a real bug?

vdfdev:
Testing using `yarn run dev` seems fine, I will try to run the same environment cypress is running...

vdfdev:
Yeah, it is a race condition. We need to wait the page to load fully, I've added a 2s delay that fixed it. There is probably a better way to do it but I will re-think our e2e tests to include minikube anyway, so this will get rebuilt.

vdfdev (approved):

PR #825, opened 2021-05-14
Remove eslint plugin to make lint fast again

vdfdev:
It seems like https://github.com/freeboardgames/FreeBoardGames.org/commit/5b44602c3ba1864bdd903ca0c4588f474c635a71 made the eslint much slower, this PR proposes to revert it. More info about hte problem here (it gets worse with more typescript files):
https://duncanleung.com/why-slow-plugin-typescript-eslint-performance-issues/

The advantage of using this plugin is that it avoids circular dependencies, but it slowed down a lot the linting.

Benchmark on my desktop (an above-average machine), on branch master it takes almost 10 minutes(!!!):
```
vinny@desktop-feliz:~/playground/FreeBoardGames.org$ time yarn run lint
yarn run v1.22.5
$ node cli/lint/lint.js
 FBG  Linting EVERYTHING... If you only care about one game, try yarn run lint GAME
$ /home/vinny/playground/FreeBoardGames.org/web/node_modules/.bin/eslint --max-warnings=0 --ext .ts,.tsx src/
$ /home/vinny/playground/FreeBoardGames.org/web/node_modules/.bin/prettier --check './**/*.{ts,tsx,js}' '../cli/**/*.js'
Checking formatting...
All matched files use Prettier code style!
$ /home/vinny/playground/FreeBoardGames.org/fbg-server/node_modules/.bin/eslint --max-warnings=0 '{src,apps,libs,test}/**/*.ts'
 FBG   SUCCESS 
Done in 383.08s.

real    6m23.243s
user    9m34.239s
sys     0m14.595s
```

On this branch, it takes less than a minute:
```
vinny@desktop-feliz:~/playground/FreeBoardGames.org$ time yarn run lint
yarn run v1.22.5
$ node cli/lint/lint.js
 FBG  Linting EVERYTHING... If you only care about one game, try yarn run lint GAME
$ /home/vinny/playground/FreeBoardGames.org/web/node_modules/.bin/eslint --max-warnings=0 --ext .ts,.tsx src/
$ /home/vinny/playground/FreeBoardGames.org/web/node_modules/.bin/prettier --check './**/*.{ts,tsx,js}' '../cli/**/*.js'
Checking formatting...
All matched files use Prettier code style!
$ /home/vinny/playground/FreeBoardGames.org/fbg-server/node_modules/.bin/eslint --max-warnings=0 '{src,apps,libs,test}/**/*.ts'
 FBG   SUCCESS 
Done in 25.06s.

real    0m25.236s
user    0m50.428s
sys     0m1.112s
```

Can we avoid circular dependencies any other way? Wouldn't we fail to compile and e2e catch it ? If not, can we make it a separate command from linting ? Linting needs to be fast to give quick feedback to developers. 

#### Checklist

### DIFF
--- a/web/.eslintrc.js
+++ b/web/.eslintrc.js
@@ -1,10 +1,9 @@
 module.exports = {
   parser: '@typescript-eslint/parser', // Specifies the ESLint parser
-  plugins: ['@typescript-eslint', 'import'],
+  plugins: ['@typescript-eslint'],
   extends: [
     'plugin:react/recommended', // Uses the recommended rules from @eslint-plugin-react
     'prettier/@typescript-eslint', // Uses eslint-config-prettier to disable ESLint rules from @typescript-eslint/eslint-plugin that would conflict with prettier
-    'plugin:import/typescript',
   ],
   parserOptions: {
     ecmaVersion: 2018, // Allows for the parsing of modern ECMAScript features
@@ -22,19 +21,10 @@ module.exports = {
     '@typescript-eslint/no-unused-vars': 'warn',
     'react/no-unused-state': 'warn',
     'no-console': 'warn',
-    'import/no-cycle': 'error',
   },
   settings: {
     react: {
       version: 'detect', // Tells eslint-plugin-react to automatically detect the version of React to use
-    },
-    'import/parsers': {
-      '@typescript-eslint/parser': ['.ts', '.tsx'],
-    },
-    'import/resolver': {
-      typescript: {
-        alwaysTryTypes: true, // always try to resolve types under `<root>@types` directory even it doesn't contain any source code, like `@types/unist`
-      },
     },
   },
 };
--- a/web/package.json
+++ b/web/package.json
@@ -175,8 +175,6 @@
     "cypress": "^5.4.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.5",
-    "eslint-import-resolver-typescript": "^2.4.0",
-    "eslint-plugin-import": "^2.22.1",
     "express": "^4.17.1",
     "graphql": "^15.3.0",
     "husky": "^6.0.0",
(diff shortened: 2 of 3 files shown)


vdfdev:
CI dropped from ~17 mins to ~ 11mins

richardaum:
It seems like a matter of preference: because IMO linting **_should_ be fast** but **_must_ have confidence**, so everything bellow is related to my opinion about this subject :heart: 

- :one: We already know cyclic dependencies are a problem  - because they _silently_ prevent the server from starting. E2E wouldn't know if it is about a cyclic dependency or not, but ESlint would know because it analyses the file structure using an AST, E2E/Cypress don't.
- :two: Linting all files is most like a _CI task_ than a _development-time task_ - because when you are coding you normally handle a couple of files, not all of them.
- :three: That is why we added a commit hook - to only lint changed/staged files. This speeds up the linting process because a linter will only fail on a file if anyone changes that file - the side effects are well-controlled and predictable on linting.
  - One could argue: "but  `--no-verify` skips the git hooks" - that is right. That is why CI should lint all files, not the developer - so "it wouldn't need to wait" for the linting output of all files, only to those changed.
- :four: Besides that, VSCode + ESLint extension do a pretty good job lint only the opened files. It also keeps a cache especially for ESlint, this way the whole process is also optimized.

TL;DR Against my opinion, maybe we could create two config files:
- `eslintrc.json`, which includes cyclic dependencies check, runs only for VSCode and commit hook.
- `eslintrc.ci.json` for CI, which does not include cyclic dependencies check; but, again, if speed is the only reason, I wouldn't follow that path.

vdfdev:
Ok then, let's leave how it is for now. I think we can get the best of both worlds if we can have the cyclic dependency done in another way, but I will investigate this later

richardaum:
@jasonharrison Any ideas about this subject? Maybe we need a third opinion.

vdfdev:
I found this tool:
https://andrejsabrickis.medium.com/locate-circular-dependencies-in-typescript-modules-2b1eb03dbf2e

Maybe it can also do the dependency check but faster. I will play with it
later today

PR #845, opened 2021-05-15
fix: get gameCode from matchId query param

richardaum:
When playing an online game, the Match page doesn't translate the game strings properly.

That is because there is no `gameCode` query param. The solution applied here was using `matchId` to get the `gameCode`.

#### Checklist


vdfdev:
Amazing! I didn't know you could do that with hooks, very powerful :D. You are starting to convert me to hooks :P

vdfdev (approved):

PR #846, opened 2021-05-16
Fix dev error for async GameProvider

vdfdev:
Fixes the error below that started popping up after https://github.com/freeboardgames/FreeBoardGames.org/pull/845
image

#### Checklist


vdfdev:
This bug is in prod too, even though it doesnt show anything to the user, so I will force submit this to fix it ASAP.

PR #863, opened 2021-05-17
Horizontally scaling bgio server

vdfdev:
This is still WIP. Somehow after setting up the redis channel, the bgio server isn't sending the messages, even if I only have one instance running.

#### Checklist


vdfdev:
Ok, I am really close. The server now works with a single instance (I was using a wrong version of socket.io-redis before). But I dont see the messages going through redis pub/sub (even though I see the channels created). I will need to investigate this further.

vdfdev:
And by deleting the "INGRESSCOOKIE" multiple times, I was able to confirm that if two clients are in different servers, they cant see the updates either

vdfdev:
FYI This PR got stuck because boardgame.io doesnt support it, I am sending PR upstream to add support, but thats going to take some time https://github.com/boardgameio/boardgame.io/pull/966

vdfdev:
I sent the code upstream and bgio 0.47.0 added support for it. Going to use that

vdfdev:
Tested with minikube and everything seems to work great ! @jasonharrison, @richardaum could you PTAL?

PR #870, opened 2021-05-18
[i18n] Use descriptive values from locales/en.json on GameDef

richardaum:
From now on, we should only use the locales files to change the description of a (partially or fully) translated game - and possibly change or request others to make the same change on other available languages for that game.

#### Checklist

### DIFF
--- a/web/src/games/chess/index.ts
+++ b/web/src/games/chess/index.ts
@@ -1,13 +1,13 @@
 const Thumbnail = require('./media/thumbnail.jpg');
 import { GameMode } from 'gamesShared/definitions/mode';
 import { IGameDef, IGameStatus, IGameTranslationStatus } from 'gamesShared/definitions/game';
-import instructions from './instructions.md';
+import translation from './locales/en.json';
 
 export const chessGameDef: IGameDef = {
   code: 'chess',
-  codes: { en: 'chess', pt: 'xadrez' },
+  codes: { pt: 'xadrez' },
   translationStatus: { pt: IGameTranslationStatus.PARTIAL },
-  name: 'Chess',
+  name: translation.name,
   contributors: ['flamecoals'],
   imageURL: Thumbnail,
   minPlayers: 2,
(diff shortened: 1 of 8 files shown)


richardaum:
@JayaDeHart you may want to check this out.

web/src/games/zooparade/locales/en.json
```diff
@@ -1,4 +1,11 @@
 {
+  "name": "Zoo Parade",
+  "description": "A cooperative game where the Team wins!",
+  "descriptionTag": "Tag for Zoo Parade TODO",
+  "instructions": {
+    "videoId": "LQ8iwNjBW_s",
+    "text": "Zoo Parade is a cooperative game for two to five players. This game perfectly combines the concepts of cooperation and partial information. The unique point: You know all cards, buy your own!\n\nThe game gets more difficult the more players you have - but at the same time this is when it is the most fun!\n\nThank you for the [Art](https://www.freepik.com/free-vector/cute-zoo-animals-card-set_4855079.htm).\n"
```
vdfdev, line 7: This is something we might need to improve in the future... It is quite hard to write a large corpus of text in a JSON field like this. We probably want to re-factor the GameInfo page to accept a filename instead and dynamically import() it there.

web/src/games/chess/index.ts
```diff
@@ -1,13 +1,13 @@
 const Thumbnail = require('./media/thumbnail.jpg');
 import { GameMode } from 'gamesShared/definitions/mode';
 import { IGameDef, IGameStatus, IGameTranslationStatus } from 'gamesShared/definitions/game';
-import instructions from './instructions.md';
+import translation from './locales/en.json';
 
 export const chessGameDef: IGameDef = {
   code: 'chess',
-  codes: { en: 'chess', pt: 'xadrez' },
```
vdfdev, line 8: Great that we dont need to repeat english anymore :)

vdfdev (approved):

PR #872, opened 2021-05-19
[i18n] Deprecate render methods used by customizations

richardaum:
The proposal here is to deprecate `renderQuick` and `renderFull`, used by customization. 

**Why?** Currently, it isn't possible to use hooks or HOCs directly inside those render methods because they are not components, only returns React elements. As an alternative, a second component that wraps everything returned by a render method would be necessary. Using a `useCurrentGameTranslation` or `withCurrentGameTranslation` is necessary to translate customization, like [this one](https://github.com/freeboardgames/FreeBoardGames.org/blob/137424c4d24f28d1e6bddce7a51975f2c47b3480/web/src/games/tictactoeplus/customization.tsx#L25).

**What is coming in?** I'm proposing that a component (or multiple) be provided as customization. The signature is pretty similar, but by letting React handle the components it allows to properly use HOC or hooks.

**What is next?** Change all places that implement a render method for customization.

#### Checklist

### DIFF
--- a/web/src/gamesShared/definitions/customization.ts
+++ b/web/src/gamesShared/definitions/customization.ts
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { ComponentType } from 'react';
 import { GameMode } from './mode';
 
 export enum CustomizationType {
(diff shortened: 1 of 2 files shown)


vdfdev:
Do we need to deprecate if first? I think we are only using the customization framework in 3 or 4 places, so if it is similar, I dont think it would be hard to just switch everything ? What do you think?

vdfdev:
oh, now I noticed that you already provided QuickCustomization interface, sounds good then :)

vdfdev (approved):

richardaum:
Great! Thanks for reviewing. As you well said, there are only 3 or 4 uses of those render methods. On my next PR, I could replace them.

PR #896, opened 2021-05-28
refactor: replace raw English text on secretDracula

richardaum:
This game was a little tricky to translate but in the end, I could finish it.

The most difficult part is because there are raw text constants and symbols. Fortunately, i18next provides a `$t` implicit function that allows users to reuse an already translated string into another translation. 

Also, I converted some text/symbol constant to there, but not all of them because some are being used inside logic - I decided to keep those there. But basically `SY_*` becomes `symbol.*`, `N_*` becomes `cname.*`.

There are some multiline messages, represented by an array in the code. These arrays of string were moved as a single string with `\n`. Then `.split('\n')` does the job of converting it back to an array.

Last thing, given that this game has A LOT of text, I would ask to review those translations carefully :pray: 

#### Checklist


vdfdev (approved):

vdfdev:
:tada: looks great! This game has a ton of text compared to others. I found some minor issues that I fixed, but now everything should be amazing :)

richardaum:
@vdfdev thanks for fixing up my PR

PR #898, opened 2021-05-28
Uses madge instead of eslint plugin to check for circular dependencies

vdfdev:
This way we can have a fast linter AND avoid circular dependencies :)

#### Checklist


richardaum (approved):

richardaum:
:rocket:

Issue #901, opened 2021-05-31
[Nine Men's Morris] Wrong Player's Piece Captured


Kurt-von-Laven:
**Describe the bug**
When the player who went second formed their first mill, the player who went first was presented with the option to choose which of their opponent's pieces to remove. The player who went second was told it was no longer their turn if they tried to select a piece to remove. When the player who went first formed mills, they were correctly able to remove the player who went second's pieces.

**To Reproduce**
1. Go to [Nine Men's Morris on web](https://www.freeboardgames.org/en/play/ninemensmorris).
2. Play with an online friend.
3. Capture each other's pieces during phase I.
4. The wrong player's piece will sometimes be captured.

**Expected behavior**
When one player forms a mill, they get to capture the other player's piece.

vdfdev:
@JosefKuchar FYI

Kurt-von-Laven:
I think I found an issue, but not the whole issue. See the linked PR for further details.

cwatsonc:
@Kurt-von-Laven is this still an issue?

Kurt-von-Laven:
Okay to close this issue, or did you want to leave it open to investigate it further?

PR #908, opened 2021-06-04
Correct bug in gameplay of Nine Men's Morris.

Kurt-von-Laven:
Don't end the placement phase if a player has to remove a piece. The consequence of this bug is that control switches players too early when a mill is formed in the final turn of the placement phase. Hence, the wrong player's piece is removed.

Improve comments for Nine Men's Morris.

#### Discussion

This pertains to #901. It is a conservative change, but I am skeptical that it fully addresses the issue since the placement phase also ends too early, and control incorrectly switches players if the 17th piece to be placed creates a mill.

#### Checklist


vdfdev:
Thank you!

vdfdev (approved):

PR #912, opened 2021-07-04
Add language selector

vdfdev:
Adds a language selector to the top bar:
image
image

#### Checklist

### DIFF
--- a/web/src/infra/common/components/base/FreeBoardGamesBar.tsx
+++ b/web/src/infra/common/components/base/FreeBoardGamesBar.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/prop-types */
 import React from 'react';
 import Select from '@material-ui/core/Select';
 import MenuItem from '@material-ui/core/MenuItem';
@@ -20,10 +19,6 @@ type FBGBarProps = {
 const handleLanguageChange = (event) => {
   window.location.href = `/${event.target.value}`;
 };
-
-function renderValue() {
-  return <LanguageIcon />;
-}
 
 function renderLanguageSelect() {
   const { i18n } = nextI18Next.useTranslation();
@@ -51,7 +46,7 @@ function renderLanguageSelect() {
           },
         }}
         onChange={handleLanguageChange}
-        renderValue={renderValue}
+        renderValue={() => <LanguageIcon />}
       >
         <MenuItem value="en">English</MenuItem>
         <MenuItem value="pt">Português</MenuItem>
--- a/web/src/infra/common/components/base/FreeBoardGamesBar.tsx
+++ b/web/src/infra/common/components/base/FreeBoardGamesBar.tsx
@@ -1,55 +1,102 @@
+/* eslint-disable react/prop-types */
 import React from 'react';
+import Select from '@material-ui/core/Select';
+import MenuItem from '@material-ui/core/MenuItem';
 import AppBar from '@material-ui/core/AppBar';
 import Toolbar from '@material-ui/core/Toolbar';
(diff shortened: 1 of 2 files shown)


richard-lopes-ifood (approved):
:rocket:

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
```diff
@@ -1,55 +1,102 @@
...
+      <Select
+        value={i18n.language}
+        className={classes.select}
+        inputProps={{
+          classes: {
+            icon: classes.icon,
+          },
+        }}
+        onChange={handleLanguageChange}
+        renderValue={renderValue}
```
jasonharrison, line 54: Is there a reason you chose to use the `renderValue` variable, as opposed to keeping it inline?  Something like:

```typescript
renderValue={<LanguageIcon/>}
```

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
```diff
@@ -1,55 +1,102 @@
...
 
-interface FBGBarProps {
+type FBGBarProps = {
   FEATURE_FLAG_readyForDesktopView?: boolean;
   toolbarContent?: React.ReactNode;
-}
+};
+
+const handleLanguageChange = (event) => {
+  window.location.href = `/${event.target.value}`;
```
jasonharrison, line 21: I recommend using NextJS's router for this instead: https://nextjs.org/docs/api-reference/next/router#routerpush

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
```diff
@@ -1,55 +1,102 @@
...
+        renderValue={renderValue}
+      >
+        <MenuItem value="en">English</MenuItem>
+        <MenuItem value="pt">Português</MenuItem>
+      </Select>
+    </div>
+  );
 }
 
+const FreeBoardGamesBar: React.FC<FBGBarProps> = (props) => {
```
jasonharrison, line 63: For what it's worth, there is a less-verbose syntax for declaring a functional component:

```typescript
function FreeBoardGamesBar({toolbarContent}: Props) {
   return <p>hai</p>
}
```

You don't need to change unless you want to

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
```diff
@@ -1,55 +1,102 @@
+/* eslint-disable react/prop-types */
```
jasonharrison, line 1: Since FBG uses TypeScript, I feel like `react/prop-types` might be a rule we want to disable project-wide.

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
vdfdev: Done. I was playing with more complex variations.

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
vdfdev: I tried that, but this wont work because the language files are loaded in the initial server render, so we need to force a full reload.

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
vdfdev: Done.

web/src/infra/common/components/base/FreeBoardGamesBar.tsx
vdfdev: I agree, done.

PR #917, opened 2021-08-04
Barnbarter

Spooky-0:
#### Checklist

#### Content

This pull request contains only the game logic, and testing. Afaik it is compelete and mimics exactly https://www.ravensburger.de/spielanleitungen/ecm/Spielanleitungen/20753%20anl%201534225.pdf?ossl=pds_text_Spielanleitung
(sorry for german xD)

#### TOOD
All of the fronted.  There is currently no frontend at all
The video linkis to one of a remastered version of the game. However I have implemented the original.


web/src/games/barnbarter/board.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+interface IBoardState {
+  hintKey: string | null;
+}
+
+export class Board extends React.Component<IBoardProps, IBoardState> {
+  constructor(props) {
+    super(props);
+    // this.state = { hintKey: null };
```
vdfdev, line 22: Remove unused comment

web/src/games/barnbarter/index.ts
```diff
@@ -0,0 +1,24 @@
+const Thumbnail = require('./media/thumbnail.jpg');
+import { GameMode } from 'gamesShared/definitions/mode';
+import { IGameDef, IGameStatus } from 'gamesShared/definitions/game';
+import instructions from './instructions.md';
+
+export const BarnBarterDef: IGameDef = {
+  code: 'barnbarter',
+  name: 'Barn Barter',
+  contributors: ['Spooky-O'],
+  minPlayers: 1,
```
vdfdev, line 10: Hummm 1 player would be local mode, right? You wouldn't need to go over the network to play. So this should be 2 and add the GameMode.Local to the modes list below

web/src/games/barnbarter/helpers.ts
```diff
@@ -0,0 +1,82 @@
+import { IG } from './definitions';
+import { Ctx } from 'boardgame.io';
+
+export function finished(G: IG, ctx: Ctx) {
+  if (G.cards.length !== 0) {
+    return false;
+  }
+
+  for (var i = 0; i < ctx.numPlayers; i++) {
+    var animals = G.players[i].cards.map((card) => {
```
vdfdev, line 10: Use let instead of var, here and eveywhre else. The scope of Var is global and we don't want that

web/src/games/barnbarter/game.test.ts
```diff
@@ -0,0 +1,340 @@
...
+  };
+  // initialize the client with your custom scenario
+
+  const clients = [0, 1, 2].map((value) => {
+    return Client({ ...spec, playerID: value.toString() } as any);
+  });
+
+  clients.map((p) => p.start());
+
+  // var { G, ctx } =  clients[0].getState();
```
vdfdev, line 32: Delete comment

web/src/games/barnbarter/game.test.ts
```diff
@@ -0,0 +1,340 @@
+import { Client } from 'boardgame.io/client';
+import { Local } from 'boardgame.io/multiplayer';
+import { BarnBarterGame, _setup } from './game';
+import { Ctx } from 'boardgames.io';
+
+// https://github.com/nicolodavis/boardgame.io/blob/master/docs/testing.md
+it('Whole Game', () => {
```
vdfdev, line 7: Would be nice to add smaller tests too, i.e. for the helper functions. This helps with debugging in case of a bug.

web/src/games/barnbarter/definitions.ts
```diff
@@ -0,0 +1,62 @@
...
+  ctx: Ctx;
+  moves: any;
+  playerID: string;
+  isActive: boolean;
+  gameArgs?: IGameArgs;
+  events?: any;
+}
+
+export interface ICard {
+  name: string;
```
vdfdev, line 19: Probably using an enum is better, since we can only have few types of animals?

vdfdev (changes requested):

vdfdev:
Some TS errors: https://github.com/freeboardgames/FreeBoardGames.org/pull/917/checks?check_run_id=3245754062#step:7:1574

vdfdev:
hey Spooky, just checking in how you are doing about this PR? Do you need any help to unblock it? Let me know how I can help!

vdfdev:
hey @Spooky-0 , just checking in how you are doing about this PR? Do you need any help to unblock it? Let me know how I can help!

Spooky-0:
I'm doing great since I am on holiday 😊😊😊
I'll be back home by the end of the week I'll take a look then

vdfdev:
Ah, great! Enjoy and take your time :)

web/src/games/barnbarter/helpers.ts
Spooky-0: yup i'm terrible at typescript. Thanks for the hint :)

web/src/games/barnbarter/definitions.ts
Spooky-0: True. But I think its fine like this. I expect no updates to the game logic, only to frontend.

web/src/games/barnbarter/game.test.ts
Spooky-0: Yes thats true. I'm not good at doing unit tests. I like the whole end-to-end, because it proves to me that all actions that the user can take work as expected. i'll add a lot of 'assertions' to make sure that if something breaks, its easy to see exactly where.

The problem with testing individual moves is that it always takes a lot of setup to reach the exact state you want to be at. E.g. a certain move is only possible very late in the game. So I just create a long long game going down all paths.

vdfdev:
@Spooky-0 I didn't see this comment until now! I will try to help you today :)

vdfdev:
At first glance in your question, I think you can pass a callback parameter as a prop to the child component, and then the parent can handle whatever it needs. I will find an example for you

cwatsonc:
@Spooky-0 I think the design goal is achieved in React by placing your state as close to the root of your components as possible -- a technique which manages state for the aggregate components you described either by: a) properties pushed down, with handlers for return communications; OR by b) contexts that have a common state that can be propagated without maintenance intense "hard wiring" (i.e. property drilling).  The place to store your state should remain high (closer to root) in this hierarchy as is possible.

vdfdev:
As @cwatsonc mentioned, an option is to use handlers for return communications (option a), which is what I used for Seabattle: https://github.com/freeboardgames/FreeBoardGames.org/blob/832bc86972981e26b035a7848ded74f135be81df/web/src/games/seabattle/Battle.tsx#L20

Option B you can read more about here: https://reactjs.org/docs/context.html ... I particularly think that A is simple and good enough for most use cases.

vdfdev:
hey @Spooky-0, did that help? Let me know if you are still stuck, I can jump into the code and help too :)

vdfdev:
This PR has been open for a long time, I fixed the CI and I am merging and leaving it as `IN_DEVELOPMENT`. Please address all comments before the next PR for publishing.

vdfdev (approved):

PR #921, opened 2021-08-09
Upgrade boardgame.io

vdfdev:
We have failing tests from mergers and secretDracula... I was trying to debug the issue with merges but got totally stuck.

It happens here: 
https://github.com/freeboardgames/FreeBoardGames.org/blob/269dae953bfc5488ea899c051623956d58be25be/web/src/games/mergers/game.test.ts#L1049
  I thought it was a  boardgame.io bug but couldnt reproduce it in this proof of concept: https://codesandbox.io/s/proof-of-concept-mergers-bug-66bp1?file=/src/Game.js


vdfdev:
@lfkellogg maybe you can help me debugging this mergers bug? I am not very familiar with the rules

vdfdev:
Good news is that https://github.com/boardgameio/boardgame.io/issues/979 seems to be the cause of our issues... bad news is that I still need to figure that out

vdfdev:
@lfkellogg fyi this seems to be an issue with the bgio framework, so no action needed from your side :)

vdfdev:
Seems like a legit bug that the e2e catched

lfkellogg:
> @lfkellogg fyi this seems to be an issue with the bgio framework, so no action needed from your side :)

Sorry I missed your ping! The fix LGTM, thanks for sorting that out.

PR #924, opened 2021-08-12
Moves setActivePlayers out of endIf to onMove

vdfdev:
This on endIf is causing a infinite loop on newer versions of BGIO

#### Checklist


vdfdev:
Merging now as I want to unblock the new version of BGIO  ASAP.

@Spooky-0 FYI

PR #930, opened 2021-08-14
Upgrade boardgame.io 0.47.0 -> 0.47.1

vdfdev:
#### Checklist


vdfdev:
Failing log: 
[mergers_fail_log.txt](https://github.com/freeboardgames/FreeBoardGames.org/files/6987229/mergers_fail_log.txt)

web/src/games/mergers/game.ts
```diff
@@ -586,23 +579,43 @@ export const MergersGame: Game<IG> = {
...
+        for (const player of Object.values(G.players)) {
+          if (G.merger.swapAndSells[player.id] === undefined) {
+            return false;
+          }
+        }
+        return true;
+      },
+
+      next: (G: IG) => {
+        if (!G.merger) {
```
tuxor1337, line 593: @vdfdev had it almost right: The `next` hook is only called after `onEnd` so that the `merger` attribute has been erased already when we come to this point. Originally, when this check was in the "next player" hook, this was before the `onEnd` call so that there was still one `mergingChain` left (even though that last chain was already dealt with).

vdfdev:
Wooohooo 🎉  thank you @tuxor1337 !!! Now we can use the latest version of bgio!

PR #933, opened 2021-08-22
New game: Picnic Go

ProspectPyxis:
#### Checklist

This is Picnic Go, intended to be a clone of the deckbuilding game Sushi Go. It's currently very heavily in progress, and I've only implemented the card logic and the main game loop.

#### TODO

* All the frontend stuff - UI, thumbnail, card art, even the instructions
* Actually implementing the rounds and win conditions properly


web/src/games/picnicGo/cards.ts
```diff
@@ -0,0 +1,145 @@
+import { cardTypes, ICardDefinition, IPlayer } from './types';
+
+export const cardDefinitions: ICardDefinition[] = [
+  {
+    id: 'sandwichChicken',
```
vdfdev, line 5: You could use an enum instead of string here, which will give you type safety. Also, you don't need to store the name in the state, you can have a function that given the Id, outputs the name.

web/src/games/picnicGo/game.ts
```diff
@@ -0,0 +1,187 @@
...
+        g.players[h.currentOwner] = cdef.scoreFunc(g.players[h.currentOwner]);
+
+        g.hands[i].hand.splice(h.selected, 1);
+        g.hands[i].selected = null;
+        g.hands[i].currentOwner = (h.currentOwner + 1) % ctx.numPlayers;
+      }
+
+      // Hands are empty, end of round
+      if (g.hands[0].hand.length === 0) {
+        // Score the chips
```
vdfdev, line 91: This function is a bit large, you could simplify it by moving the scoring logic to a sub function for instance

vdfdev (approved):
Great work! I just left a few comments, feel free to merge after addressing or ignore them if you don't agree

web/src/games/picnicGo/cards.ts
ProspectPyxis: Just overhauled most of the card system - it now uses enums and gives the name in a separate function just as you suggested. As a bonus, I also got rid of the deck property in the global game object, seeing as it's not used for anything.

web/src/games/picnicGo/game.ts
ProspectPyxis: Done.

vdfdev:
That looks great, thank you!

PR #935, opened 2021-08-23
Picnic Go: backend code completed

ProspectPyxis:
#### Checklist

This should complete the game logic and every cards' functions - potato chips and cakes now score properly, forks can now be used as intended to take two cards, and the game now returns a scoreboard on game end. Only thing left now is the frontend - UI, card graphics, so on.


vdfdev:
Looks great, thank you! By the way, before publishing we need to add unit tests at least to cover the rules (it is a requirement so we can make sure the game doesnt break in the future when upgrading libraries, for instance). Let me know if you need any help with that. There are examples on the codebase and some documentation here on how to do it : https://boardgame.io/documentation/#/testing

vdfdev (approved):

PR #948, opened 2021-09-07
Removes google analytics

vdfdev:
#### Checklist


jasonharrison (approved):
I'm not sure how I personally feel about removing Google analytics (how do we track games played offline?) but the code changes look right to me.  Do what you think is best with GA/analytics in general

vdfdev:
I agree it isn't an ideal state after this PR, but analytics is also broken after i18n for per-game stats (it was only tracking English). So for now I think the online-only dashboard is enough, and I will follow up on a discussion with the community how to get the missing metrics (like non-online-mode usage)

PR #953, opened 2021-09-12
Update Schafkopf rules

tuxor1337:
This fixes a little UI problem and adds a "called player runs away" rule to the Schafkopf game for 4 players: There are no "official" rules for this game, but typically, there are restrictions in which situations the called player is allowed to play the called suit (i.e., "run away"), especially the Ace.

There are more rule variants for Schafkopf, especially for the round scoring, but I felt like this one is the most important missing from the current implementation. Furthermore, the UI problem made the game unplayable in some situations (with a React error message on the screen).

#### Checklist

### DIFF
--- a/web/src/games/schafkopf/components/GameBoard.tsx
+++ b/web/src/games/schafkopf/components/GameBoard.tsx
@@ -104,7 +104,7 @@ export function Board(props: {
     return (
       <Trick
         trick={trick ? trick.cards : []}
-        leaderPos={trick ? +trick.leader.id : 0}
+        leaderPos={trick && trick.leader ? +trick.leader.id : 0}
         winnerPos={trick && trick.winner ? +trick.winner.id : -1}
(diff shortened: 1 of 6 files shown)


web/src/games/schafkopf/components/GameBoard.tsx
```diff
@@ -104,7 +104,7 @@ export function Board(props: {
     return (
       <Trick
         trick={trick ? trick.cards : []}
-        leaderPos={trick ? +trick.leader.id : 0}
+        leaderPos={trick && trick.leader ? +trick.leader.id : 0}
```
tuxor1337, line 107: Before this PR, this line could lead to an error message `trick.leader is undefined`.

vdfdev: Thanks for pointing out :)

vdfdev (approved):

Issue #678, opened 2020-11-20
[Windows Dev] Bgio server isn't starting if using the CLI


vdfdev:
Developers on windows get  "Failed to start a Match" when trying to start a Match on windows.

Seems to be an issue with the CLI, unknown reason yet. The current workaround is to do  "yarn run gen:games && yarn run dev:internal" on the root folder to bypass CLI.

flamecoals:
I tried on my windows machine, and the issue seems to be with "start-server-and-test" probably not liking the env created by shelljs....

I want to in the future manage the instantiation and running of the 3 servers on the CLI itself, so we can present a better UI for the user. This would hopefully solve this issue.

sefgit:
What are the different ports number for ?
3000: web server
3001: proxy ?
8001: bgio ?

Tried to find them in documentation but without luck.

vdfdev:
Fixed by #955

PR #955, opened 2021-09-15
Replace mkdir and cp in copy.ts with native Node calls.

mateusazis:
`mkdir` and `cp` do not work on Windows as-is because they take different formats there. For instance, `mkdir` on Windows uses `-Force` instead of `-p`. Using Node calls should make the script OS-agnostic.

#### Checklist

### DIFF
--- a/web/server/config/i18n/sync/copy.ts
+++ b/web/server/config/i18n/sync/copy.ts
@@ -1,10 +1,14 @@
-import { execSync } from 'child_process';
 import { regexp } from './constants';
+import { copyFileSync, existsSync, mkdirSync } from 'fs';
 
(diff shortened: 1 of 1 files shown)


vdfdev:
Nice! I did not know that! Maybe this will fix https://github.com/freeboardgames/FreeBoardGames.org/issues/678 ?

vdfdev (approved):

mateusazis:
`yarn run dev checkers` works fine on Windows for me now (despite some errors about i18n). It failed before that.

I am not 100% sure if this fixes that issue though. I remember I also had some issues with `run-p` not being found which I fixed by installing it globally (`npm i -g npm-run-all`).

PR #956, opened 2021-09-16
Recalculate pre-selected token on checkers state change to account for AI moves.

mateusazis:
Previously, I was pre-selecting a coordinate on `_move`. It turns out this is only called in the React world by drag and click events. AI calls moves directly. Thus, `componentDidUpdate` is a better place for this calculation.

Preview: https://drive.google.com/file/d/105i5haOXtGZW30yc4EmwGr6iAxjyeCEJ/view?usp=sharing

#### Checklist


vdfdev (approved):

vdfdev:
Looks great, thank you!

Issue #941, opened 2021-08-27
[Request for comments] Rethinking public rooms to foster a player community


vdfdev:
# The problems

1) **Our public rooms features are underutilized** - Even though it tooks us a lot of effort to develop, it never "catched on" with players. See the numbers below:
```
fbgDb=> SELECT "isPublic", COUNT(*) FROM room_entity GROUP BY 1;
 f        | 607430
 t        |  18308
```

Out of all the rooms created, only 2.9% of the rooms were public.

2) **Users that come to the home page directly might not realize there is a private room option** --- See https://github.com/freeboardgames/FreeBoardGames.org/issues/784. I've noticed that I've been often kicked out of public rooms, probably because they were not meant to really be public.

3) **Our community of developers is disconnected with the community of players.** Our discord is mainly inhabited by the developer community (which is great!) but we don't have much player activity there. Seeing other people play your game is very motivating for me, and even though we have https://stats.freeboardgames.org showing thousands of users, we don't have much personal connection with the players.

# Proposals

## Short-term proposal

Move the carousel to select a public room out of the [home page](https://www.freeboardgames.org/en/) and into the [game info](https://www.freeboardgames.org/en/play/checkers), sitting below the game mode cards. This should still be a carousel with several mini-game cards, and NOT look like the existing game mode cards, because the public room might be unrelated to the game the user is looking into. However, we should rank the public rooms in a way that always shows the game of the page first.

### Pros
- More traffic seeing public rooms, more likely to find an opponent after creating room (improving 1)
- Less confusion about private x public, as both options would be in side by side (improving 2)
- Might allow users to find other games

### Cons
- Might confuse users to see rooms for other games that are not the game they are looking into.
- The game page info is latency sensitive as the main entrypoint for users, so we need to make sure there is no layout shift (like it has today on the home), and also does not increase a lot the Lighthouse metrics.
- Does not tackle issue 3

## Long-term proposal

We should replace the carousel and the public room with a simpler matchmaking experience. The idea is as follow:

### v1
- Remove the carousel completely and replace it with a separate game mode, let's call it "Ranked".
- Users who click on the "Ranked" game mode are going to be asked to log-in with Discord.
- After that, a spinner will ask the user to wait for X minutes (X = 2?). It will also have a link to a new channel on discord #matchmaking. If someone else does the same during this time we start the match, otherwise ask the user to try again.
[...]

PR #960, opened 2021-09-19
Creates public room webhook and adds discord link on room

vdfdev:
Several improvements to try to foster a player community as discussed on #941:
- Link to discord on our room page

image

-  Notifies #lets-play when someone opens a public room

Screenshot from 2021-09-19 00-14-05

- Reduces from 60 mins to 5 mins how long a public room shows up after inactivity, reducing the "ghost room" problem.

#### Checklist


ProspectPyxis:
I like this change! It's a simple way to increase interaction with our Discord server without outright requiring players to log in to play.

A couple quick questions/points on this, though:
- Does the pop-up go away permanently if you click the x, or is it just for the current room and it'll come back the next time? I ask this because I'm not sure if needing to close the pop-up every time you just want to create a private game with friends might get too annoying or not - it might be something to keep an eye on so that it doesn't start feeling *too* intrusive.
- Should the pop-up only show up for the game's host, maybe? Presumably, they'd be the one deciding if they'd like more players to join in with their game, not the players joining in.

These are just suggestions, not must-implements - I'd like to hear what you think.

vdfdev:
For your first point, no, it is not permanent when you click on the x. I hope that it isnt too intrusive if the user doesnt want to close the reminder, but I agree with you ideally it would hide for a few days at least. 

For both of your points, my main concern right now is that there wont be enough players on the #lets-play channel, so I am leaning toward erring on the side of showing it too much than too little (for hosts vs guest, or for hiding forever or not). Also, I think this solution is the simplest for now, let's see how users behave with these changes and then if we get enough players, we can try to make it a bit more intelligent. Or if we don't get enough players at all, we can give up on these reminders.

fbg-server/src/rooms/rooms.service.ts
```diff
@@ -35,37 +37,41 @@ export class RoomsService {
...
     if (!queryRunner) {
       await inTransaction(this.connection, async (queryRunner) => {
-        await this.saveNewRoom(queryRunner, userId, roomEntity);
+        await this.saveNewRoom(queryRunner, user, roomEntity);
       });
       if (room.isPublic) {
         // Cant do this inside a transaction as we need the new room in the notification.
         await this.lobbyService.notifyLobbyUpdate();
+        const doNothing = () => { return; };
+        this.maybeNotifyDiscord(user, roomEntity).then(doNothing, doNothing); // Fire and forget
```
jasonharrison, line 49: I'm just curious, why is the .then(doNothing, doNothing) necessary? What's the difference between this and ignoring the returned promise?

jasonharrison (approved):

fbg-server/src/rooms/rooms.service.ts
vdfdev: You were right, this is not necessary.

ProspectPyxis:
Definitely agree with your points. I've sat on this for a bit, and I now definitely see merit in showing the link too much rather than too little. No further complaints from me, in that case!

PR #961, opened 2021-09-19
Add Doppelkopf card game

tuxor1337:
This implements the popular German card game [Doppelkopf](https://en.wikipedia.org/wiki/Doppelkopf). As it uses a subset of the cards that are used by Skat, it doesn't come with it's own card artwork included, but imports the artwork from the Skat game directory.

Includes English, German, French and Portuguese (please check those!) locales, as well as unit tests.

This does only implement the "official" ruleset by the DDV (German Doppelkopf Association). This game is particularly known for having a vast range of regional variants. 

In addition to the implementation of Doppelkopf, this PR also applies some code rearrangements in my other card games and adds some rule variants to the Schafkopf game. If you feel that this is too much for one PR, I will be happy to open a separate PR for that.

The next games I have on my to do list are [Zwanzigerrufen](https://en.wikipedia.org/wiki/Zwanzigerrufen) and [Königrufen](https://en.wikipedia.org/wiki/K%C3%B6nigrufen) (both popular in Austria).

One more comment: By putting all my added games at the top of the list of games on the front page of FreeBoardGames.org, a user might think that this project is focused on card games. It's not my intention to "capture" or "dominate" FBG, and I love the diversity of games we have. Please feel free to divert from the rule (ordering by recency added) and rearrange the games in any order you prefer.

#### Checklist


vdfdev:
Amazing work again @tuxor1337. I tested on both mobile and desktop and it works well (I need to catch up learning the rules though, hehehe).

I found a bug when randomly clicking on some buttons:
image

Another issue is that on Desktop this button seems to be overlapping with the player name:
image


I will see if I can fix it for you. The portuguese translation has some minor improvements that I will push but overall LGTM, we can launch!

And regarding the game order... Don't worry about it. From time to time I re-order the games that have been launched ago accordingly to their popularity (as shown in stats.freeboardgames.org ...). I will do that soon with the older games you already contributed. You are in no way capturing/dominating or anything like that, your work is very much apprieciated ! If you want to keep sending games or also help reviewing code and working on the infra-structure, you are very much welcome ! We are in this to have fun and this is a hobby for us :D

vdfdev:
I am not entirely sure why `Annoucement[A]` is giving us problem. It seems to be valid TS: https://stackoverflow.com/questions/18111657/how-to-get-names-of-enum-entries  Maybe A is undefined ?

I will leave this PR pre-approved, you can merge after your check that you liked the fix for the button alignment bug, and maybe fixed that other bug?

vdfdev (approved):

tuxor1337:
Thanks for your review and thanks fixing the button positioning! I fixed the problem with the announcement buttons. This was very well spotted indeed :)

PR #962, opened 2021-09-19
Picnic Go: First iteration of UI complete

ProspectPyxis:
#### Checklist

This PR finishes up the UI, meaning that clicking on the cards and buttons now work as they should. And of course, players should now be able to see and play the game without using the debug panel.

With this update, the game is pretty much in a fully playable state - however, there is still some polishing to be done before it's ready for public release:
- The UI as of now is completely untested for mobile. While it should work in theory with the way I've set everything up, actually seeing it in action might be another story.
- Certain UI elements may still need some polish - in particular, I'd like to figure out if I can order the way played cards are displayed so it's easier on the players needing to keep track of how cards interact.
- Unit tests for the board still need to be written. I think I get the gist of how to do this from studying code from other games, but don't be surprised if I start needing to ask for help in discord over this, haha.
- Migrate every bit of raw text into i18n, to future-proof the game for localization. From what I understand so far, this is going to require me to rewrite all the class components into function components, but other than that, this shouldn't be too difficult to do.
- I'd like some way to tell the players what each of the cards do - every card has been redesigned from the game this is cloning from, so it may be difficult for players to tell which card is equivalent to which without a guide. I'm still trying to figure out what the best way to convey this information would be, though.
- And of course, making the thumbnail image and writing the instruction text.


vdfdev:
Code looks great, I am going to playtest a bit now :)

vdfdev:
The UI looks great! Amazing job :D. Very simple and functional :).

I tested on mobile and it played well! The only thing blocking launching is writing the instructions and creating a thumbnail. 

Optionally, you might also consider using this component: https://www.freeboardgames.org/docs/?path=/story/infrastructure-common-carousel-carousel--mobile On desktop it automatically detects that it goes beyond the allowed width and shows a button to the user (some users might miss the horizontal bar). But we might need to adapt it to have different height. This is completely optional, just a NIT.

We can merge now and fix these blockers in a follow up PR :) Great job again!

vdfdev (approved):

vdfdev:
Oh, another thing that would be great to do before launching... Instead of hard-coding all strings, we could use the translation function instead. This way, it is super easier for translators to translate the game to other languages. I can translate to portuguese for you :)

ProspectPyxis:
Thank you for your review! Yes, I'm definitely planning on migrating everything to translation - in fact, this is actually already done on my local environment, haha.

Other than that, then, I'm still figuring out the best way to tell the player information on how each card scores, and also polishing up a few UI points that slightly bothers me still. This shouldn't take too long, and you should be seeing a fully publishable version of Picnic Go very soon!

PR #963, opened 2021-09-21
Make Scoreboard component extensible

ProspectPyxis:
#### Checklist

This allows the `Scoreboard` component to be extended with as many additional columns as necessary per game. I've also implemented a visible example with storyboard. Hopefully, it shouldn't be too hard to figure out what this is supposed to do based on the example I provided.

Feel free to merge if everything looks good as it is, but I do have two points I'd like to raise:
- Right now, all the extra components are listed between the player name and the score. Should this be moved after the score instead?
- There's currently no check to ensure that `extraData` and `extraColumns` are the same length. I admit that I don't know how to make react throw an error properly here, but I'm not sure how necessary that check is, since the columns would obviously look misaligned anyways if there's a length mismatch.

### DIFF
--- a/web/src/gamesShared/components/scores/Scoreboard.stories.tsx
+++ b/web/src/gamesShared/components/scores/Scoreboard.stories.tsx
@@ -10,6 +10,14 @@ const scores = [
   { playerID: '2', score: 300 },
 ];
 
+const scoresExtended = [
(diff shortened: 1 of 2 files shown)


vdfdev (approved):

vdfdev:
Thank you!

PR #954, opened 2021-09-15
Pre-select board tokens in checkers when there is only 1 valid move.

mateusazis:
This should make it easier for new players who might not understand why some tokens cannot be selected. It also makes the game a bit faster.

Preview: https://drive.google.com/file/d/1-sPZN4MuLXD0XObg0e8i22RZvXoFqdrt/view?usp=sharing
_(please ignore the eternity it takes me to plan my moves)_

#### Checklist

* [X] Use a separate branch in your local repo (not `master`).
* [X] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).

Issue #946, opened 2021-09-04
[Checkers] Possible game improvements/new game options


ProspectPyxis:
**Is your feature request related to a problem? Please describe.**
I played a game of checkers with a friend, and I noticed quite a few flaws that make the game not as smooth as it could be. Plus, I also had some thoughts about possibly adding new configuration options to the game.

**Describe the solution you'd like**
I'd like to see the following features implemented in the game:
- A configuration option to enable or disable forced capture.
- Better indication for invalid moves when there is a forced capture on the board - for example, a red flash on the piece you're trying to move.
- A configuration option for whether kings can move any amount of squares or only one square per turn. 
- Might be a bigger scope than the other two configuration options, but an option to set the amount of pieces each player has between 8 (2 rows of pieces) and 12 (3 rows).

The rationale behind the configuration options is that checkers seems to be one of those games where everyone plays it differently in some way, and it can be quite jarring to be forced into one standard ruleset - my friend was very surprised about kings going any distance diagonally, while I was surprised that the game uses three rows of pieces per player instead of two.

As for the request for making invalid moves with forced capture more clear, the rationale is that it can be very jarring to have moves simply quietly fail with no indication, as I've run into the problem of not seeing the forced capture and being perplexed that my move just seems to not register at all a few times. I'm not sure how viable this is code-wise, though.

vdfdev:
I completely agree, and this is something that would add a lot of value to FBG because it is our most played game! I will mark this bug as high priority and a good first issue (it isn't too hard)

ProspectPyxis:
Going to list #604 as a related issue - it seems my friend isn't alone in being confused at the king's movement.

vdfdev:
This PR https://github.com/freeboardgames/FreeBoardGames.org/pull/954  will happy alleviate some of those issues! Thanks @mateusazis for the great contribution!

ProspectPyxis:
Every point I've brought up in this issue (and then some!) should be closed by #971!

Issue #604, opened 2020-06-20
[Checkers] Bug with logic that determines valid moves for kings


eemp:
It looks like there may be a bug with checkers logic determining valid moves for kings unless the following is intentional:

![image](https://user-images.githubusercontent.com/4442378/85212046-ce4f9680-b31c-11ea-874f-0f5944da5853.png)

flamecoals:
Thanks for reporting this! I will have a look into the rules and see if this was intentional or not

flamecoals:
@JosefKuchar was this intentional or not ?

As a kid I've played like this (jumping the whole diagonal), but I think the most common rule played in US just go one space backward.

flamecoals:
FYI, I believe this might be a misunderstanding about checkers rules, as they vary by region.  I think we should change the rule to be based on the most common US/International played rules, which would make this move illegal. However, I also believe we need to open the possibility of rule variations within the game.

flamecoals:
Checkers is our most played game right now, and we are showing up as the first positions in search engines for it. So I will raise this as high priority because it might be surprising/upsetting people.

ProspectPyxis:
Closed by #971, there's now a configuration option to enable or disable "flying kings" - it's also off by default.

vdfdev:
Great job @ProspectPyxis, thank you for fixing this long-standing issue!!!

PR #971, opened 2021-09-24
Checkers: A lot of general improvements, new customization settings

ProspectPyxis:
#### Checklist

This adds, well, a lot of improvements to checkers, the biggest change being the new customization interface. If fully implemented, this should fully close both #604 and #946.

Please read through the list of changes carefully, as I've went ahead and made some changes that may be controversial - I can revert any changes that is believed to be a bad idea.

#### Changes Made:
- Rewrote all of `board.tsx` into a functional component, rather than a class component. React hooks are used to replicate all the functions of a class component's state, so functionally, this should be practically identical to before.
- Migrated all text to use the translate function. This is made possible by the functional component rewrite above. Please note two things about this change:
  - Despite the game already having locale files for `de` and `pt`, they are very much incomplete, practically not there at all - the only things translated are the game name and description. I know neither language, unfortunately, so I'm leaving it up to others to fill in the blanks.
  - I've rewritten the instruction text somewhat, the biggest change being that I added a new paragraph describing the game's settings being changeable in various ways. If you think this should be reverted, please let me know.
- Added *five* configuration options that directly affect the gameplay. I've also changed some defaults, so the game now plays a little differently from how it was before even if you don't touch the settings at all. As it stands, the ruleset should now closely imitate English Draughts by default, but let me know if any of the defaults should be changed around. The settings available are:
  - **Pieces Per Player** - set between 12 (default) or 8. I'll admit that this is a bit of self-indulgence on my part, as in my region, checkers are only played with 8 pieces per player, and I'd like to be able to play that variant on this site as well.
  - **Forced Capture** - on by default. Should be self-explanatory.
  - **Kings Can Move Any Distance** - off by default (used to be "enabled"). Also known as flying kings, this should be self-explanatory as well.
  - **Stop Jumping If Piece Kinged** - on by default (used to be "disabled"). If this option is on, a jump sequence immediately ends if the jumping piece turns into a king, so it can't continue capturing backwards.
  - **Draw After n Moves Without Progress** - set between off, 15, 30, or 40 (default). This is an entirely new rule that adds a draw condition to the game, to prevent games lasting forever due to, say, both players having one king each and just constantly jumping around avoiding each other without making a capture. In essence, this acts like chess's 50-move rule - if the selected amount of moves pass without either player making a capture or a non-king move, the game is automatically a draw.
- If forced capture is on and there are forced moves on the board, the selectable pieces are now highlighted cyan. This is basically a companion to #954 - players should be able to tell pieces they can select now even if there's more than one capture available on the board.

About the only thing left I haven't done is a better AI than randomly selected moves - I did give it a shot, but my minimax implementation was quite slow even at low depths, and I couldn't tell if that was an issue with my own code, if there's some existing unoptimized code in the game logic, or if it's just that slow normally, so in the end, I scrapped it and decided to make this PR as is. It's still something to look into at a later date, though.

### DIFF
--- a/web/src/games/checkers/board.tsx
+++ b/web/src/games/checkers/board.tsx
@@ -173,12 +173,16 @@ export function Board(props: IBoardProps) {
       if (isFirstPersonView(props.gameArgs, props.playerID)) {
         if (winner === props.playerID) {
           return 'you won';
+        } else if (winner === 'draw') {
+          return 'draw';
         } else {
           return 'you lost';
         }
       } else {
         if (winner === '0') {
           return 'white won';
+        } else if (winner === 'draw') {
+          return 'draw';
         } else {
           return 'black won';
         }
--- a/web/src/games/checkers/board.tsx
+++ b/web/src/games/checkers/board.tsx
@@ -30,82 +30,66 @@ interface IBoardProps {
   gameArgs?: IGameArgs;
 }
 
-interface IBoardState {
-  selected: ICartesianCoords;
-  validMoves: IMove[];
-}
-
 function roundCoords(coords: ICartesianCoords) {
   return { x: Math.round(coords.x), y: Math.round(coords.y) };
 }
 
-export class Board extends React.Component<IBoardProps, IBoardState> {
-  state: IBoardState = {
-    selected: null,
-    validMoves: getValidMoves(this.props.G, this.props.ctx.currentPlayer),
-  };
+export function Board(props: IBoardProps) {
+  const [selected, setSelected] = React.useState(null);
+  const [validMoves, setValidMoves] = React.useState(getValidMoves(props.G, props.ctx.currentPlayer));
 
-  isInverted() {
-    return (isAIGame(this.props.gameArgs) || isOnlineGame(this.props.gameArgs)) && this.props.playerID === '1';
+  function isInverted() {
+    return (isAIGame(props.gameArgs) || isOnlineGame(props.gameArgs)) && props.playerID === '1';
   }
 
-  _isSelectable = (coords: ICartesianCoords) => {
-    if (isOnlineGame(this.props.gameArgs) && this.props.playerID !== this.props.ctx.currentPlayer) {
+  const _isSelectable = (coords: ICartesianCoords) => {
+    if (isOnlineGame(props.gameArgs) && props.playerID !== props.ctx.currentPlayer) {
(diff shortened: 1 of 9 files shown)


vdfdev:
Nice! I can help you translating to Portuguese. I am just about to start work, I can have a look at this tonight :)

web/src/games/checkers/board.tsx
```diff
@@ -30,106 +32,106 @@ interface IBoardProps {
...
 function roundCoords(coords: ICartesianCoords) {
   return { x: Math.round(coords.x), y: Math.round(coords.y) };
 }
 
-export class Board extends React.Component<IBoardProps, IBoardState> {
-  state: IBoardState = {
-    selected: null,
-    validMoves: getValidMoves(this.props.G, this.props.ctx.currentPlayer),
-  };
+export function Board(props: IBoardProps) {
```
vdfdev, line 39: Thanks for converting this to functional component!

web/src/games/checkers/game.test.ts
```diff
@@ -15,8 +18,21 @@ test('invalid moves', () => {
...
 it('should declare player 1 as the winner', () => {
+  const CheckersGameWithSetup = {
+    ...CheckersGame,
+    setup: () => ({
+      board: INITIAL_BOARD[0],
+      jumping: null,
+      config: {
+        ...DEFAULT_FULL_CUSTOMIZATION,
+        flyingKings: true,
+        stopJumpOnKing: false,
```
vdfdev, line 29: Ideally we would add a bit more of coverage here, for these new configs.

ProspectPyxis: I agree - the only reason I didn't is that these config options are the bare minimum to allow the already written full game test to work. I'd really prefer not to rewrite every move by hand all over again with the new configs, so maybe a better way would be more general unit tests?

vdfdev: Oh, I think that maybe playing a single game for each config would be enough. No need to rewrite everything. We can merge this PR and follow up later. Everything LGTM.

vdfdev:
LGTM, this improves a lot Checkers, our most played game, thank you @ProspectPyxis !!!

vdfdev (approved):

ProspectPyxis:
Remember to close the linked issues once this is merged!

PR #973, opened 2021-09-27
Checkers: Minor improvements, new unit tests

ProspectPyxis:
#### Checklist

This PR does a few big things:
- The `G` object has been optimized a little - instead of storing the entire board, it now only stores where all the pieces are. This should functionally behave identically to before, but it should slightly reduce both the amount of looping required and the amount of data sent back and forth between server and client.
- There's now a new function that will create a board from a string, and `INITIAL_BOARD` now reads from these strings rather than a full array. This, once again, should be functionally identical to before, but it allows for...
- New unit tests have been added. The new `convertStringToBoard` function allows for testing from a much wider variety of positions without needing to manually define where each piece is in the board array, and I've leveraged this to create new unit tests like what we discussed in the previous checkers PR.

I'll be honest, I actually made these changes in anticipation of implementing a minimax-based AI, and I wanted to optimize the game code to reduce the time it'd need to spend processing. However, it quickly grew a little too big, so I've decided to spin it off into its own thing - hopefully you'll see the payoff some time, haha.

### DIFF
--- a/web/src/games/checkers/board.tsx
+++ b/web/src/games/checkers/board.tsx
@@ -131,31 +131,28 @@ export function Board(props: IBoardProps) {
   }
 
   const getPieces = () => {
-    return props.G.board
-      .map((piece, index) => ({ data: piece, index }))
-      .filter((piece) => piece.data !== null)
-      .map((piece) => {
-        const { x, y } = fromPosition(piece.index);
-        return (
-          <Token
-            x={x}
-            y={y}
-            draggable={true}
-            shouldDrag={_shouldDrag}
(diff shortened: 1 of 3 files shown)


vdfdev:
This looks great! Thanks for the refactor, just going to test a bit and merge

vdfdev:
Played a game, worked great!

I just found a minor issue: After you select a piece, you cant un-select it. However, this issue seems to be in prod already, so we can merge this PR and follow up with a fix later.

vdfdev (approved):

PR #976, opened 2021-09-28
Checkers: Bugfix, piece deselecting

ProspectPyxis:
#### Checklist

This PR does two things:
- Fixed a bug with the game end logic. I've also added a new unit test that should check for this in the future.
  <details>
    <summary>Details of the bug</summary>
    Consider this position:

    image

    After white takes the forced capture, white technically has no moves left (it's up against the edge, and its only square forward is blocked by two pieces). However, it should now be black's move, which means the game shouldn't end quite yet. With the former logic using `endIf`, the game would incorrectly end immediately after this move, since the game detects that white no longer has any moves remaining, but doesn't account for whose turn it is after the move.

    To fix this, I've simply moved the valid moves check to `onBegin`, and added a safety return to `board.tsx` to handle an empty valid move array, if it ever gets passed.
  </details>
- Pieces can now be deselected by simply clicking any other non-move square.

This should be the last bit of refactoring I'm doing with checkers for the time being if no new bugs are found - I've decided that implementing a better AI is something I wish to leave for the future.

### DIFF
--- a/web/src/games/checkers/board.tsx
+++ b/web/src/games/checkers/board.tsx
@@ -59,7 +59,11 @@ export function Board(props: IBoardProps) {
     if (selected === null && _isSelectable(position)) {
       setSelected(position);
     } else {
-      _move(position);
+      if (validMoves.some((move) => equals(move.to, position))) {
+        _move(position);
+      } else {
+        setSelected(_isSelectable(position) ? position : null);
+      }
     }
   };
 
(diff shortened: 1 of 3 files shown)


vdfdev:
Great job finding and fixing this corner case issue with the game end logic, and fixing the unselecting bug!

vdfdev (approved):

PR #985, opened 2021-10-13
Schafkopf 3 player mode with short deck

tuxor1337:
#### Checklist

This replaces the previous 3-player variant of Schafkopf by a more common variant where the card values 7 and 8 are removed from the game (a "short" deck) and there is no kitty. Furthermore, the layout of the Card component is fixed so that the card shadow is not cropped.


vdfdev:
LGTM, but need to fix this compiler error before merging:

```


./src/games/schafkopf/tests/gen.ts:51:12
Type error: Expected 1 arguments, but got 0.

  49 |   const handSize = 8;
  50 |   const cmpCards = util.get_cmpCards(G.contract, CardColor.Herz);
> 51 |   G.deck = getSortedDeck();

```

PR #987, opened 2021-10-14
Notifications

### LINKED ISSUE #61: [Infrastructure] Automatic notifications for the user when it is their turn so they can play.

richardaum:
This PR _probably_ closes #61

#### Checklist

| Untouched | Mute notifications | Unmute notifications
| -- | -- | -- |
image | image | image

#### Notes
- By default, the notifications are enabled after browser's first time acceptance.
- A mute/unmute option is added to every game from GameLayout component.
- The config for every game is modified on Game component to listen for turns change.
- There is a lot of conditions to satisfy after being able to show notifications: permission granted on browser, isNotAlreadyNotified, isYourTurn, isScreenActive, isAllowedGameMode, is not muted.
- Notifications and menu are translated only to english and portuguese (as these are the languages I feel comfortable to translate).
- Due to inability of rerendering Game component on local games (not sure if it is only there), I needed to combine a Context API - to update the menu labels to mute/unmute - with a `staticContext` to actually mute and unmute 😞 

#### Side quests
- Refactor `GameLayout` to a functional component.
- Wrap App on Game component to isolate logic and be able to modify the game config isolately.


vdfdev (changes requested):

cwatsonc:
> > The code looks great, and this will be a great improvement to FBG :D. I tried to test this locally, but could not trigger the notification both on firefox and chromium. I did not get any prompt from the browser asking if I should get notifications (maybe bc I am running on localhost ?). To try to debug it, I put a breakpoint on canBeNotified() and on the notify() functions, but these dont seem to be called at any times.
> > Do you know what might be happening?
> 
> Not sure. I tried to run the code again and suddenly `turn.onBegin` callback is not being called anymore - and it is a callback from bg.io which triggers the entire logic 😞

Not sure how much this impact the solution but some games may not track turn as the moment of passing actionability -- this is true in cardtable that uses "turn" to encapsulate the "hand" in cribbage games -- play doesn't pass in a steady rhythm as the various contexts arise.  Wouldn't it be better to give the game a hook that it could invoke when appropriate rather than blanket turn based notifications?  @vdfdev @richardaum

cwatsonc:
With the work of 10/29/2021 where contributors @richardaum @cwatsonc met and [discussed in this video](https://vimeo.com/640571466/c150001071) we see multiple behaviors that qualify as edge cases that are sub-states of the "happy - path" 

1.  that browser/notification state can cause unexpected behaviors (multiple locations starting at 1:55) there is a lot of discussion leading to state is a _conjunction of server definitive game state and the browsers' individual "notification" states_ -- the browsers maintain their own individual states based upon os prefs, and user prefs that are mapped both to the browser _per site_ and global preferences for the notification preferences. [need references here to times] 
2. a graceful handling of error conditions must be captured by the client code as demonstrated at (citations required)
3. the end-of-game discussion that causes the users to reregister and opt-in could be improved (citation required)
4. the per game opt-in for notifications should be discussed and mitigated (citation required)
5. addition requirements found during play (feel free to add comments below and we can discuss as a team)

richardaum:
I've change a few wrappers that were breaking some tests, that is why the CI was failing. Now, it is working. Everything looks good so far.

web/src/games/tictactoe/index.ts
```diff
@@ -26,6 +27,7 @@ export const tictactoeGameDef: IGameDef = {
   config: () => import('./config'),
   aiConfig: () => import('./ai'),
   customization: () => import('./customization'),
+  notifications: new Set([Notification.TurnChanged]),
```
vdfdev, line 30: Do we really need to customize this per-game? I think as a user I would like the consistent notification behavior across all games... I.e. notify me when it is my time to play.

Also, we are not checking this for showing the menu option... Seems to be adding complexity with little gain. I think we should remove this option from IGameDef

cwatsonc: This is noted in the video and believe @richardaum gave a rationale -- best to check with him on the "why"

richardaum: As we have games that do not follow the turn rule, I'm not sure how notifications would be useful on those games. Or even worse - if notifications disturb players on those game non-turn based. If you guys agree, I can remove the `notifications` from game def.

vdfdev (approved):

PR #1004, opened 2021-11-06
Add english pattern to shared Card component

tuxor1337:
This adds a full 52-cards deck in Anglo-American ("English") pattern to the shared `<Card />` component. This PR replaces the current PNG artwork used in the `cardtable` game (by @cwatsonc) with the new SVG-based pattern.

The artwork's license is public domain: https://commons.wikimedia.org/wiki/File:English_pattern_playing_cards_deck_PLUS.svg

#### Checklist


tuxor1337:
There are some issues with the tiled SVG image that is included as a CSS `background-image` property. For some reason, the image works fine in storybook, but doesn't work inside of games:

* In storybook, the path to the SVG file in the CSS `background-image` property is mostly untouched: `static/media/src/gamesShared/components/cards/media/cards_english.svg`
* Inside of games, the path to the SVG file in the CSS `background-image` property will be replaced by `_next/static/images/cards_english-6af167bd803550a3975f08510fe07de1.svg` and the file will contain something like this: `export default "data:image/svg+xml;base64,PD94bWwgd...` Clearly, the browser can't handle this as an SVG file.

tuxor1337:
I'm wondering: What is the reason for applying the following rule to SVG files, currently?
https://github.com/freeboardgames/FreeBoardGames.org/blob/70deda5cd4f3f98d62e08590f6bc0855bccce9eb/web/server/next.config.ts#L58-L66

web/server/next.config.ts
```diff
@@ -56,7 +56,7 @@ export default withWorkers(
       });
 
       config.module.rules.push({
-        test: /\.(webp|svg|mp3|wav)$/,
+        test: /\.(webp|mp3|wav)$/,
         exclude: /node_modules/,
         use: [
           {
```
tuxor1337, line 62: Is there a particular reason why we use the `url-loader` for SVG files, currently? Here, I remove SVG files from this rule because otherwise, I can't url-import SVG files in CSS `background-image` properties.

vdfdev: At some point in time I tried to use SVGs directly at seabattle, but latter I gave up and wrapped them in react components, so I think this is fine

vdfdev (approved):

PR #1010, opened 2021-11-18
Upgrade typescript to 4.5

ProspectPyxis:
#### Checklist

This is the start of a long-haul effort to bring as many of our dependencies up to date as possible.

As of now, only typescript will be upgraded. Other dependencies will come in later PRs.

### DIFF
--- a/web/src/games/chess/ai.ts
+++ b/web/src/games/chess/ai.ts
@@ -17,7 +17,7 @@ class Stockfish {
   worker: any;
 
   constructor() {
-    this.worker = new (StockfishWorker as any)();
+    this.worker = new StockfishWorker();
   }
 
   start() {
--- a/web/src/games/chess/ai.ts
+++ b/web/src/games/chess/ai.ts
@@ -17,7 +17,7 @@ class Stockfish {
   worker: any;
 
   constructor() {
-    this.worker = new StockfishWorker();
+    this.worker = new (StockfishWorker as any)();
   }
 
   start() {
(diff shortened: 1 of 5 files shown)


cwatsonc:
Question:  the focus of this PR is typescript?  Can we modify the linter to provide a warning of outdated typescript as an initial step? We can then start the process of migrating between the two specifications. Again making this a long running PR seems off-target.  The longer it is open the more maintenance it will add as MASTER is improved.  Suggest we rethink this to fit in surgical steps /$0.02

web/src/games/chess/ai.ts
```diff
@@ -17,7 +17,7 @@ class Stockfish {
   worker: any;
 
   constructor() {
-    this.worker = new StockfishWorker();
```
richardaum, line 20: Is it possible to create a d.ts file containing a simple definition of `StockfishWorker` rather than casting to `any`?

ProspectPyxis: I'll need to look into this - I'd imagine it would be, but I don't know how quite yet.

ProspectPyxis:
> Question: the focus of this PR is typescript? Can we modify the linter to provide a warning of outdated typescript as an initial step? We can then start the process of migrating between the two specifications. Again making this a long running PR seems off-target. The longer it is open the more maintenance it will add as MASTER is improved. Suggest we rethink this to fit in surgical steps /$0.02

That's probably a good call - I'll restrict this PR to just updating Typescript for now.

vdfdev (approved):

vdfdev:
Thank you!

Issue #989, opened 2021-10-20
[Documentation] Provide a baseline infrastructure description for development


cwatsonc:
Description: The fbg documentation doesn't currently prescribe minimum infra required to develop/build the solution.  The designers have chosen monorepo and choices for improvements are driven by currently supported platform and versions.

Suggestion: Provide a baseline support document describing our platform so developers may align their efforts with the community and ease adoption of our framework.

Proposed:
Provide a platform minimal description that is supported in the current build.  Annotate the package.json with properties that which will warn/fail process if not present.  If yarn is our chosen build tool, create a minimum version which is supported to give guidance on tools choice. Add an "engines" property in package.json to enforce/warn minimums at build time. Add "cpu" and "os" descriptions with similar prescriptive advice. Make these mutable to allow speculative development but matching should guarantee a successful build and easy integration of contributions by development. 

Alternatives Considered:
I have been reading the yarn docs and looking forward to the yarn 2.x.x upgrades -- I have been reading multiple articles on alternatives for build orchestration (e.g. bit, npm).  Yarn 2 is a radical shift in build behavior and structure but promises swift operational builds on first download.  I will compile a list of articles discussing the rationales.

Additional Context:
[yarn 2 migration and rationales](https://yarnpkg.com/getting-started/migration)
[Alternative Package Managers](https://blog.bitsrc.io/4-npm-alternatives-best-js-package-managers-and-publishing-tools-fe6779937ee9)
[Yarn manifest docs](https://yarnpkg.com/configuration/manifest)
[Engines Plugin for Yarn](https://github.com/devoto13/yarn-plugin-engines)

cwatsonc:
created PR #1013

richardaum:
> Yarn 2 is a radical shift in build behavior and structure

Yes, it is, definetely. And I think in that case the costs overcome the benefits. I mean, Yarn 1 is doing a good job here, right?

cwatsonc:
@richardaum the issue is discussed on the "rationales" link on the main issue -- yarn@1.0x.x is about to be orphaned for support -- the yarn team is poised to update to 3.x.x in the next 6 months and will ensure that fact.  

[...]

PR #1013, opened 2021-11-20
Feature/documents infra

cwatsonc:
#### Checklist

* [x ] Use a separate branch in your local repo (not `master`).
* [ x] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-contributing--page).

this PR provides an anchor to Infra documentation as per #989 
makes refinements to the structural organization of docs


web/docs/03_infra/1.Overview.stories.mdx
```diff
@@ -0,0 +1,30 @@
...
+
+## Core Infra
+
+The following table describes a core elements of FBG Infra:
+
+###todo: break down further between server, web, build, packaging
+
+| What | Description        | Type |
+| -------- | ------------------ | ------- |
+| Node.js       | ES Runtime for executing js   | Native Libary |
```
richardaum, line 20: I think it is better to give a link to the library's docs instead of than describing them here.

richardaum (approved):

web/docs/03_infra/1.Overview.stories.mdx
cwatsonc: Agreed I think I'll add that as any preamble to the support documents - that being said, there is room for introducing HOW we use these products -- most are peer dependencies (Node, Yarn, etc) and don't always show in meaningful ways in our code artifacts.  I also think this needs additional refinement into 4 or 5 sub categories: Web, Server, Build, Deploy, etc. @richardaum

vdfdev:
@cwatsonc feel free to merge

PR #1014, opened 2021-11-20
Add UserReport to collect feedback

### LINKED ISSUE #1006: User feedback about its experience
### Context

Currently, we have the following ways we can receive feedbacks from our users:
- By submitting an issue here on Github.
- By joining our discord server and chatting with us.
Both of those require some effort - which could make the user give up on giving a _maybe_ good feeback.

### Hypothesis

If we provide a faster way to the players give a feedback about their experiences, it could encourage them to actually send a feeback.

### Features

- An in-game form to submit feebacks (e.g. rating with or without comments - suggestiong, bug reports, etc.)
- Automatic attaching a screenshot of its screen (only displaying the game, preserving its identity, hiding chat content, etc.)
- Automatic integration with Github by creating an issue with some specific label.
- Provide the Github issue link to the user (if he/she would like to follow the created issue).

### What could go wrong?

- Spamming
- Poor feedback

richardaum:
This PR closes #1006. 

It adds UserReport as a tool to collect feedback. This tool adds a button on our webiste where user can send feeback about its experieence. 

All received feeback will be recorded on UserReport - right now, only I have access to the tool. So let me know if you want to also receive the feedbacks.

**Preview**

image

image

image

image

#### Checklist

### DIFF
--- a/web/src/pages/_app.tsx
+++ b/web/src/pages/_app.tsx
@@ -89,6 +89,13 @@ class DefaultApp extends App {
           <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
           <meta name="msapplication-TileColor" content="#ffc40d" />
           <meta name="msapplication-config" content="/static/icons/browserconfig.xml" />
+          {/* @ts-ignore */}
+          <meta name="userreport:mediaId" value="121cd588-61a0-48af-bc02-12aa0db7c0a9" />
+          <script
+            id="userreport-launcher-script"
+            src="https://sak.userreport.com/freeboardgames/launcher.js"
+            async
+          ></script>
         </Head>
         <ThemeProvider>
           <SelfXSSWarning />


vdfdev:
Isn't that blue button going to show on top of games and etc? I think it might mess up with the game play of some games on mobile ?

cwatsonc:
I think there's a couple reasons we might want to slow down on a full release to PRD with this tooling:

1. we should consider alternatives that might integrate better with GitHub 
2. we need to consider a strategy that places this in a sandbox deployment so dev can reflect on usability and other concerns -- the link to game requests on Discord we added a few months back comes to mind -- we need to consider how this gets used and possible harmful ways it can be weaponized against our community -- I'm down with moving forward -- let's provide a minimum of caution before going LIVE in PRD.

richardaum:
> Isn't that blue button going to show on top of games and etc? I think it might mess up with the game play of some games on mobile ?

Yeah... that isn't good. Fortunetely they have an API we can write our own button. What about creating a about on top bar?

richardaum:
@cwatsonc 
> We should consider alternatives that might integrate better with GitHub.

Unless we are going to use a paid with a GH integration, I couldn't find any free tool with a Github integration. So I think is the best we can get for free.

> we need to consider a strategy that places this in a sandbox deployment so dev can reflect on usability and other concerns
We don't have a sandbox environment.

> we need to consider how this gets used and possible harmful ways it can be weaponized against our community -- I'm down with moving forward -- let's provide a minimum of caution before going LIVE in PRD.
Please be clear why you are agains moving forward on this, as we could get some really useful feedback from our players. I don't see any clear reason, besides what @vdfdev pointed out, for not trying it.

cwatsonc:
Since I haven't seen it in action I'm not certain fully how it might be exploited -- we had a recent event where the game notification announcements were used to deliver racist tropes to Discord -- I think similar behavior could be expected of the feedback tool -- not sure what the use cases are yet but I expect freeform text to be available for feedback and that might be a vector for weaponizing the tool -- I'm not against user feedback I just want us to use caution when deploying to production any improvements that might have harmful side effects. /$0.02

richardaum:
Ok, thanks for your feedback.

ProspectPyxis:
Honestly, personally, I don't believe this will be anywhere as bad as the discord webhook deal. Not to say that we won't get junk, because we definitely will, but at least with user reports, ideally it should all be private and sent to some backend rather than being sent to some public front-facing area like our discord.

vdfdev:
@richardaum I would be okay with merging this as long as it is inside the options menu, but ideally we would avoid the extra JS/dependency with third-party app, as this will increase the payload sent to browsers.

I will close this for now as there hasnt been a ton of activity in the past few months, feel free to send another PR addressing these issues

PR #1020, opened 2021-12-18
Issue 1019 | fix investigate

Spooky-0:
The problem was that someone put `message: any = ...` and later a nested object could either be a string or an array.
Now this is still possible but we at least handle both cases :)


vdfdev:
CI is still failing, whenever it is green I can merge

vdfdev (approved):

PR #1023, opened 2022-01-06
picnicGo: use Math.random as seed

tuxor1337:
@vdfdev, would you please deploy this on Freeboardgames.org to see whether the statistics change somehow? It shouldn't be noticable for the user at all.

#### Checklist

### DIFF
--- a/web/src/games/picnicGo/game.ts
+++ b/web/src/games/picnicGo/game.ts
@@ -200,7 +200,6 @@ export function confirmScore(g: IG, ctx: Ctx) {
 
 export const PicnicGoGame = {
   name: 'picnicGo',
-
   setup: (ctx) => {
     const baseState = {
       players: new Array(ctx.numPlayers).fill({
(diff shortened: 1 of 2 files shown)


web/src/games/picnicGo/game.ts
```diff
@@ -200,7 +200,9 @@ export function confirmScore(g: IG, ctx: Ctx) {
 
 export const PicnicGoGame = {
   name: 'picnicGo',
-
+  // the statistical distribution of cards on Freeboardgames.org seems to be strange,
+  // so test a different seeding approach (instead of the BGIO default based on Date.now):
+  seed: Math.random(),
```
delucis, line 205: This will seed _all_ matches with a single seed, which is probably not what you want. In an FBG context this is actually tricky to do probably as you'd need a game factory to generate the game object each time a match is created.

tuxor1337: Ah, you are correct. I fixed it in a467d9a

vdfdev (approved):

PR #1046, opened 2022-03-13
translate(fourinarow) to PT

leocaseiro:
#### Checklist

Translate Four in a Row to Portuguese!

I also made a few improvements on the Translation docs, such as including the missing languages, and linking to the i18next docs.


web/docs/02_how_to/1.GameTranslation.stories.mdx
```diff
@@ -14,7 +14,10 @@ The following table describe all languages acceptable:
 
 | Language | Description        |
 | -------- | ------------------ |
-| en       | English            |
+| en       | English (Default)  |
```
leocaseiro, line 17: Should we set English (Description) or each language should have in each language?

web/src/games/fourinarow/board.tsx
```diff
@@ -150,4 +158,5 @@ export class Board extends React.Component<IBoardProps, {}> {
   }
 }
 
-export default Board;
+const enhance = compose<IBoardProps & IBoardInnerProps, {}>(withCurrentGameTranslation);
```
leocaseiro, line 161:
```suggestion
const enhance = compose<IBoardInnerProps, IBoardProps>(withCurrentGameTranslation);
```

web/docs/02_how_to/1.GameTranslation.stories.mdx
vdfdev: I think it is ok to expect developers to use English

vdfdev (approved):

PR #1067, opened 2022-03-23
Explicitly define the type for checker's move functions.

mateusazis:
Another attempt to catch incorrect calls at compile time.

Also removed the `step` property since it was unused.

#### Checklist


vdfdev (approved):

vdfdev (approved):

vdfdev:
Ugh, GitHub action outage this morning means that the CI is stuck... Not sure how to trigger again, I think I will need a bogus commit on master to force a rebase and run again

PR #1073, opened 2022-03-28
Create dependabot.yml

vdfdev:
#### Checklist


leocaseiro (commented):
Will this update all npm packages daily?

vdfdev:
Yep! That's the idea. I still haven't enabled it to the web folder because I still need to catch up the dependencies there, but will do after that.

PR #1181, opened 2022-04-21
Disable swear filter for private matches

ProspectPyxis:
#### Checklist

Exactly what it says on the tin: this completely disables the swear filter for private matches while still having it work in public games.

Manual testing indicates this works for both the room and match screens, but we're still missing automated tests.

### DIFF
--- a/fbg-server/src/chat/chat.service.ts
+++ b/fbg-server/src/chat/chat.service.ts
@@ -12,20 +12,20 @@ const MAX_MESSAGE_LENGTH = 280;
 
 @Injectable()
 export class ChatService {
-  constructor (
+  constructor(
     private roomsService: RoomsService,
     private matchService: MatchService,
     private usersService: UsersService,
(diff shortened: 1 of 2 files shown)


fbg-server/src/chat/chat.service.ts
```diff
@@ -74,13 +80,26 @@ export class ChatService {
...
         `You need to be a player in this match in order to send messages.`,
         HttpStatus.BAD_REQUEST,
       );
     }
   }
+
+  private async checkChannelIsPublic(channelType: ChannelType, channelId: string) {
+    let roomEntity: RoomEntity;
+    if (channelType === 'match') {
+      const matchEntity = await this.matchService.getMatchEntity(channelId);
```
vdfdev, line 96: We have to double check that the match is populated from the room in this SQL query

vdfdev:
Looks great! I will just double check what I mentioned this weekend and merge it

PR #1215, opened 2022-05-02
Add new game Contract Bridge

tuxor1337:
This adds an implementation of the 4-player trick-taking card game Contract Bridge. There are several variants of that game around, most notably Rubber Bridge and Duplicate Bridge. However, this implementation is only similar to Rubber Bridge, and not exactly like any of the more popular variants. Maybe some Bridge fan out there has the time to look into the details later, and implement the correct Rubber Bridge scoring, and maybe also a "mod" for other variants like Duplicate Bridge or Chicago Bridge. For now, this implementation at least provides the basic game mechanics of the Bridge family of card games.

#### Checklist


vdfdev (approved):

vdfdev:
Looks great! Great job again @tuxor1337

PR #1239, opened 2022-05-07
Implement player ordering/shuffling in room view

tuxor1337:
This implements reordering of players in game rooms. Players might want to be in the match in a certain order (or in random order). Currently, users are always listed in the order they joined the room in the first place. The only way to change the ordering with the current implementation is to kick users out and let them join again in the right order, shuffling is difficult.

This PR adds a "Move user up" and a "Shuffle users" button to the `ListPlayers` element, only available to the creator of the room. The order is realized by adding a `position` column in the membership database. Previously, memberships were sorted by `id`. Alternatively, the creator can also choose to shuffle the ordering immediately before the match starts, so that users don't know the ordering in advance.

grafik

#### Checklist

### DIFF
--- a/common/gql/schema.gql
+++ b/common/gql/schema.gql
@@ -13,6 +13,7 @@ type NewUser {
 
 type RoomMembership {
   isCreator: Boolean!
+  position: Float!
   user: User!
 }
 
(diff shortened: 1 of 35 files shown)


vdfdev (approved):

vdfdev:
Looks great, thank you!!!

PR #2166, opened 2023-04-24
Fix typo in German locale for chess

hermannloose:
#### Checklist

* [x] I've read and agree with the [contribution guidelines](https://www.freeboardgames.org/docs/?path=/story/documentation-getting-started-contributing--page).


hermannloose:
I noticed that #2051 already proposes the same thing but has been stale for while.

